PR #333, opened 2024-01-17
Overhaul how character encoding is handled. 

### LINKED ISSUE #186: Add option to just wing it with encoding
In #118 we changed over to only using ASCII encoding. Though in the wild many problems use other encoding schemes. We should either add an option to not use ASCII, which defaults to UTF-8, or to allow the user to specify whatever encoding they want to use.

MicahGale:
Adds:
- `MCNP_InputFile` to documentation site.
- Added encoding options when opening a file, and to `input_reader`
- Added script utility to strip out all non-ascii characters.
    this works by:
    - reading into UTF-8 with replace.
    - Replaced replaced characters with nothing or `' '`
    - Covert back to ascii. Anything about 128 also gets converted. 
 - Added utilities section to guide
 - Added FAQ that shows a typical stack trace

Needs still:

- [x] test for changes
- [x] graceful error handling for bad encoding
- [x] Document how to use this in user docs.
- [x] Add FAQ: what are encoding errors
- [x] Add doc strings

Closes #186, #334.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,8 @@ jobs:
       - run: pip install .
       - run: pip uninstall -y montepy
       - run: pip install --user dist/*.whl
+        # run scripts
+      - run: change_to_ascii -h
       - run: pip uninstall -y montepy
       - run: pip install --user dist/*.tar.gz
       - run: pip install --user montepy[test]
--- a/doc/source/api/montepy.input_parser.input_file.rst
+++ b/doc/source/api/montepy.input_parser.input_file.rst
@@ -1,0 +1,9 @@
+montepy.input\_parser.input\_file module
+========================================
+
+
+.. automodule:: montepy.input_parser.input_file
+   :members:
+   :undoc-members:
+   :show-inheritance:
+   :private-members: _convert_to_int, _convert_to_enum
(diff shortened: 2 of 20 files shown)


MicahGale:
Actually how hard should we try with enforcing a valid encoding @tjlaboss? It should get pushed down to python really quick and handled that way.

MicahGale:
I'm not sure I like the command: `change_to_ascii`

MicahGale:
This also leads to the question: should we go back and support UTF-8 as the default? I am learning that ASCII is very annoying.

MicahGale:
Artifacts that include the updated website: https://github.com/idaholab/MontePy/actions/runs/7585620985?pr=333

MicahGale:
We changed the behavior from strictly enforcing encoding to replacing all non-ascii characters with spaces. This is the most conservative approach. Using default python `replace` would have lead to a lot of `?` which would break syntax. Extra spaces randomly throughout should generally have no impact on the semantic meaning of a model. If a user has a problem with formatting being messed up they can always just delete the offending non-ascii chars.

MicahGale:
Quick thought: does the character stripper handle `\r\n` properly? There are reports that mixing and matching line ends in a file will lead to complications.

tjlaboss:
Should be able to review this week.

MicahGale:
Should a warning be issued every time a character is converted?

tjlaboss:
No. If it is important to warn about specific character conversions, I'd put them all in a single warning message.

tjlaboss (approved):
The first official MontePy script!

MicahGale:
Thanks for the review. I'm going to wait until merging #336.

PR #360, opened 2024-02-04
Fixed geometry Tree with actual type enforcement

### LINKED ISSUE #351: Lack of data enforcement in syntax node is showing up again
**Describe the bug**

While working with a somewhat complicated model I wanted to see the exact MCNP input so I tried `cell._tree.format()` (I know rude, but I'll open another issue about this). This then lead to the error:

``` python
In [17]: cell._tree.format()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[17], line 1
----> 1 cell._tree.format()

File montepy/input_parser/syntax_node.py:185, in SyntaxNode.format(self)
    183             ret += node.format()
    184     else:
--> 185         ret += node.format()
    186 return ret

File /montepy/input_parser/syntax_node.py:239, in GeometryTree.format(self)
    237 ret = ""
    238 for node in self.nodes.values():
--> 239     ret += node.format()
    240 return ret

AttributeError: 'dict' object has no attribute 'format'
```

This is extremely similar to #204. 

I think it's finally time to get to the bottom of things and stop generating dictionaries instead of `GeometryTree`

**To Reproduce**
MWE of input:

```
100103  3926  5.00000E-02 (-102002          -36425  36426):   $ comment
                          (-102002  169239  -36426  36427)    u=424 
```

**Expected behavior**
No error.

**Additional context**
Version: `0.3.0dev1`

MicahGale:
The error here was that a syntax tree was being passed an invalid data type, and it was mostly hobbling along. So I enforced the data types and then actually fixed the data type issue.

fixes #351.


tjlaboss (approved):
LGTM

PR #336, opened 2024-01-22
Improving CI/CD for Github actions

MicahGale:
Improvements: 

* Moved completely to pyproject.toml
   * deleted pytest.ini
   * delete `requirements/*`
* Added coveralls support
* Added a test report summary
* Went completely to git based `setuptools-scm` for controlling version
  * version system is robust and should never lead to an error
  * versions are determined by a git tag
  * Git tag is automatically created with an action that uses semantic versioning to decide if it's a patch, minor, or major release.
 * Added badges
 * Improved deploy process.
    * implemented actual build-deploy process
    * build will:
        1. bump to a new version, and create a git tag
        2. build
        3. sign the packages with sig store
        4. Create a draft release on GitHub
        5. Upload artifacts for later deployment to PyPi
 * Created a Pull Request Template

TODO:

* [x] enable coveralls for this repo
* [x] Add coveralls as a collaborator
* [x] document how semantic versioning works
* [x] Finish implementing #326. 
* [x] Fix #328.
* [x] Fix #338. 

resolves #326, resolves #202, resolves #329, resolves #328, resolves #338.

PR #368, opened 2024-02-06
Fix bug that can't parse materials with implicit library specification. 

### LINKED ISSUE #365: Material Specifications: cannot unpack non-iterable ValueNode object
@MicahGale How are material cards supposed to be formatted for MontePy to parse them? I keep getting the error in the subject line. When I inspect the `isotope_fractions` variable in `\data_inputs\material.py`, it is a `ListNode` object containing `ValueNode` objects. For any `ValueNode` object, I see no attribute that corresponds to the isotopic fraction (at least not one intuitively named). Additionally this object is non-iterable so the usage isn't correct either. Am I missing something?

MicahGale:
The material parser was built assuming that all isotope specifications required a library, e.g., `1001.80c`. However according to the 6.2 manual table 3-32: 

> Either a full ZZZAAA.abx or partial ZZZAAA element or nuclide
identifier for each constituent i, where ...

Therefore, this assumption is flawed.

To fix this, a dedicated `MaterialParser` was created. `1001` is ambiguous if it is part of a number sequence or a ZAID. So this parser just accepts both Isotopes and number sequences as valid. This avoids the need of contextual parser with a look-ahead capability by making the input contextual as an input level. 

@SterlingButters you can review this if you like, but you don't need to. 

As for the specific error message that was found in #365. The Material assumed it had an `IsotopesNode` which is a list of tuples, and was given a `ListNode` instead. It tried to unpack the `ValueNode` inside the list like a tuple, and freaked out. 

Fixes #365.


MicahGale:
This needs to wait until #336 is merged due to the de facto deployment freeze depending on that.

tjlaboss (approved):

MicahGale:
Thank you!

PR #385, opened 2024-03-05
updated coverage options to make coverall paths work.

MicahGale:
This is a quick fix for coveralls.


MicahGale:
Bypassing review as this is a small CI PR that I verified is making Coveralls happy.

PR #398, opened 2024-04-12
Fix bug with how thermal scattering law data is not written to file

### LINKED ISSUE #395: Thermal Scattering laws get lost to the ether
**Describe the bug**

`ThermalScatteringLaw`s never get printed out to file with `problem.write_to_file`.

This is due to an older approach, where they are never loaded in `problem.data_inputs` and their parent `material` object is responsible for getting them to print.

**To Reproduce**

A short code snippet of what you have ran. Please change or remove any specific values or anything that can't be public. For example:

``` python
problem = montepy.read_input("foo.imcnp")
problem.write_to_file("bar.imcnp")
```
Example file:

```
test
C cells
1 0 -1

C surfaces
1 so 5

C data
m2001
       1001.00c     2
       8016.00c     1
mt2001 h-h2o.40t
```

f If it includes any specific values please change or remove them. For example:

```
1 1 20
         -1000  $ dollar comment
        imp:n,p=1 U=350 trcl=5

C surfaces
1000 SO 1

C data
C materials
C UO2 5 atpt enriched
m1        92235.80c           5 &
92238.80c          95
```

**Version**

 - Version 0.2.5

**Additional context**

MicahGale:
# Description

Fixes #395 

# Checklist

- [ ] ~~I have made corresponding changes to the documentation (if applicable)~~

### DIFF
--- a/montepy/data_inputs/material.py
+++ b/montepy/data_inputs/material.py
@@ -116,6 +116,21 @@ class Material(data_input.DataInputAbstract, Numbered_MCNP_Object):
                 if cell.material == self:
                     yield cell
 
+    def format_for_mcnp_input(self, mcnp_version):
+        """
+        Creates a string representation of this MCNP_Object that can be
+        written to file.
+
+        :param mcnp_version: The tuple for the MCNP version that must be exported to.
+        :type mcnp_version: tuple
+        :return: a list of strings for the lines that this input will occupy.
+        :rtype: list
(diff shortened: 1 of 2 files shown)


tjlaboss (approved):
Tested the test. Confirmed the bug existed before the fix. Tested the fix. Confirmed it is fixed.

LGTM

PR #439, opened 2024-07-09
Pass changelog-test when no code is modified

tjlaboss:
# Description

Add some additional logic to check if any code in `montepy/` was updated. 

Todo: squash

Fixes #436 

# Checklist

- [x] I have tested the CI.

PR #443, opened 2024-07-11
Implemented overwrite guardrails thanks to @tjlaboss.

### LINKED ISSUE #442: Guardrail for MCNP_Problem.write_to_file()
New function signature: 

```python
def write_to_file(self, new_problem, overwrite=False):
    """
    :param overwrite: Whether to overwrite the file at 'new_problem' if it exists
    :type overwrite: bool
    """
    . . .
```

Check for the existence of the file `new_problem`, and do not overwrite it unless `overwrite=True`.

One solution: simply raise an error.

```python
if overwrite or not os.path.exists(new_problem):
    # write the file as normal
    return
elif os.path.isfile(new_problem):
    raise FileExistsError(new_problem)
elif os.path.isdir(new_problem):
    raise IsDirectoryError(new_problem):
. . .
```

Or, if the file exists and `overwrite == False`, back up the original file with some sort of identifier, and then write to the requested filename. 

UUID:

```python
def get_backup_fname(new_problem):
    base_name, extension = os.path.splitext(new_problem)
    unique_id = uuid.uuid4()
    new_fname = f"{base_name}_backup_{unique_id}{extension}"
    return new_fname
```

Timestamp:

```python
def get_next_backup_fname(new_problem):
    base_name, extension = os.path.splitext(new_problem)
    timestamp = int(time.time())   # OR use strftime('%Y%m%d%H%M%S') for human-readable
    new_fname = f"{base_name}_backup_{timestamp}{extension}"

    # If user going faster than one file per second
    index = 1
    while os.path.exists(new_file):
        new_file = f"{base_name}_backup_{timestamp}_{index}{extension}"
        index += 1

    return new_fname
```

**Additional context**
Related to pyOpenSci review here: https://github.com/pyOpenSci/software-submission/issues/205#issuecomment-2218837269

MicahGale:
# Description

This introduces a new argument to `MCNP_Problem.write_to_file`, (and `montepy.input_parser.input_file.write_to_file.open`), `overwrite`, which indicates if the existing file should be overwritten. The default behavior is to raise a `FileExistsError` if the file exists, and overwrite is not specified. This error is ignored in `overwrite=True`. No matter what a `IsADirectoryError` is raised if the path is a folder.

Thanks @tjlaboss  for the prototyping.

Fixes #442

# Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -1,11 +1,15 @@
 MontePy Changelog
 =================
 
-0.2.10
-----------------------
-**Miscellanious Changes**
+#Next Version#
+---------------------
+
+**Features Added**
 
 * ``overwrite`` argument added to `MCNP_Problem.write_to_file` to ensure files are only overwritten if the user really wants to do so (:pull:`443`).
+
+0.2.10
+----------------------
 
 **Bug fixes**
 
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -3,6 +3,9 @@ MontePy Changelog
 
 0.2.10
 ----------------------
+**Miscellanious Changes**
+
+* ``overwrite`` argument added to `MCNP_Problem.write_to_file` to ensure files are only overwritten if the user really wants to do so (:pull:`443`).
 
 **Bug fixes**
 
--- a/doc/source/starting.rst
+++ b/doc/source/starting.rst
@@ -79,8 +79,12 @@ state as a valid MCNP input file.
 
 >>> problem.write_to_file("bar.imcnp")
 
+The :func:`~montepy.mcnp_problem.MCNP_Problem.write_to_file` method does take an optional argument: ``overwrite``. 
+By default if the file exists, it will not be overwritten and an error will be raised.
+This can be changed by ``overwrite=True``.
+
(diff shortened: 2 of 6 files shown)


MicahGale:
Do you think this change counts as a minor or patch version change?

MicahGale:
I need to update the getting started guide accordingly.

tjlaboss:
> Do you think this change counts as a minor or patch version change?
>
>> MINOR version when you add functionality in a backward compatible manner
>> PATCH version when you make backward compatible bug fixes

This fits the description of MINOR version more closely than PATCH version I'd say.

tjlaboss:
Nice to see that #439 is useful.

MicahGale:
> > Do you think this change counts as a minor or patch version change?
> > > MINOR version when you add functionality in a backward compatible manner
> > > PATCH version when you make backward compatible bug fixes
> 
> This fits the description of MINOR version more closely than PATCH version I'd say.

Ok I'll update docs to say version `0.3.0`. Yay we are off the rails of running into the version `0.2.15` train on test-pypi

tjlaboss (approved):

Issue #201, opened 2023-12-21
End of Life for python 3.8 support


MicahGale:
Python 3.8 security [end of life](https://peps.python.org/pep-0569/#lifespan) is scheduled for October 2024. MontePy should move past 3.8 by that point.

This should have minimal impact on code, and should only required metadata updates, and removing old tests. 

This does bring up a question of should we jump from almost dead version to almost dead version or keep in the middle of the pack for python support.

PR #444, opened 2024-07-15
Updated dependency to sphinx 7.4.0, and enabled linkcheck testing.

### LINKED ISSUE #414: Use linktest as part of website build
TIL: [sphinx supports linkcheck](https://www.sphinx-doc.org/en/master/usage/builders/index.html) that tests all external links. I think this should be part of the documentation tests.

MicahGale:
# Description

An upstream bug with Sphinx and linkcheck failing for PDFs with anchors has been resolved. This just requires a version of Sphinx with this fix, and requires linkcheck to pass. 

Fixes #414

# Checklist

- [ ] ~I have added tests that prove my fix is effective or that my feature works (if applicable)~

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,7 +30,7 @@ jobs:
           ref: main
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.12
       - run: pip install --user . montepy[doc]
       - run: cd doc && make html
       - uses: actions/configure-pages@v4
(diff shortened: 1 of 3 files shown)


MicahGale:
The issue is actually that sphinx 7.4.0 doesn't support python 3.8. I think I will just update build testing, and docs deploy to use a later version of python.

tjlaboss:
Python 3.8 is due to reach its scheduled upstream end-of-life in October 2024.

MicahGale:
> Python 3.8 is due to reach its scheduled upstream end-of-life in October 2024.

Yep see #201. IMO: all supported versions of python will be supported for users, but only the latest version of python will be officially supported for dev. This change would move dev support from 3.8 to 3.9. I think this is fine as building docs is a "dev" task.

tjlaboss (approved):

PR #474, opened 2024-08-08
Simplify isotope printing with __repr__

### LINKED ISSUE #473: Simplify Isotope repr
Let's say we have an O-16 `Isotope` in the 01c xslib. Currently,

```python
>>> str(o16)
' O-16  (01c)'
>>> repr(o16)
'ZAID=8016, Z=8, A=16, element=oxygen, library=01c'
```

`__repr__` is mostly used for debugging and logging purposes and should remain simple.

Move the current `__repr__` to `__str__`, and do the following instead:

```python
>>> repr(o16)
'Isotope(8016.01c)'
```

I think the `Isotope({self.ZAID})` is best, because ZAID is the most widely used convention in MCNP [citation needed]. However, the new MCNP6.3.1 style would be acceptable as well:

```python
>>> repr(o16)
'Isotope(O-16.01c)'
```

Single or double quotes around the identifier may be preferable as well: `Isotope(". . .")` or `Isotope('. . .')`

**Additional context**
More broadly, the `__repr__` functions should be simplified throughout MontePy to be brief and to show the class name. I'm targeting this one first because of #472.

tjlaboss:
# Description

* New, simpler `Isotope.__repr__`
* Add method `Isotope.nuclide_str()`
* Update _test_material.py_ accordingly

Fixes #473 

# Checklist

- [x] I have updated the changelog
- [x] I have updated tests that prove my fix is effective or that my feature works

### DIFF
--- a/montepy/data_inputs/isotope.py
+++ b/montepy/data_inputs/isotope.py
@@ -178,7 +178,8 @@ class Isotope:
         :returns: a string that can be used in MCNP
         :rtype: str
         """
-        return f"{self.ZAID}.{self.library}"
+        suffix = f".{self.library}" if self._library else ""
+        return f"{self.ZAID}{suffix}"
 
     def nuclide_str(self):
         suffix = f".{self._library}" if self._library else ""
--- a/montepy/data_inputs/isotope.py
+++ b/montepy/data_inputs/isotope.py
@@ -167,7 +167,7 @@ class Isotope:
         self._library = library
 
     def __repr__(self):
-        return f"{self.__class__.__name__}({repr(self.mcnp_str())})"
+        return f"{self.__class__.__name__}({repr(self.nuclide_str())})"
 
     def mcnp_str(self):
         """
@@ -179,6 +179,9 @@ class Isotope:
         :rtype: str
         """
         return f"{self.ZAID}.{self.library}"
+
+    def nuclide_str(self):
+        return f"{self.element.symbol}-{self.A}.{self._library}"
 
     def get_base_zaid(self):
         """
--- a/montepy/data_inputs/isotope.py
+++ b/montepy/data_inputs/isotope.py
@@ -196,7 +196,8 @@ class Isotope:
         return self.Z * 1000 + self.A
 
     def __str__(self):
-        return f"ZAID={self.ZAID}, Z={self.Z}, A={self.A}, element={self.element}, library={self.library}"
+        suffix = f" ({self._library})" if self._library else ""
+        return f"{self.element.symbol:>2}-{self.A:<3}{suffix}"
 
     def __hash__(self):
         return hash(self._ZAID)
--- a/montepy/data_inputs/isotope.py
+++ b/montepy/data_inputs/isotope.py
@@ -166,8 +166,8 @@ class Isotope:
             raise TypeError("library must be a string")
         self._library = library
 
-    def __str__(self):
-        return f"{self.element.symbol:>2}-{self.A:<3} ({self._library})"
+    def __repr__(self):
+        return f"{self.__class__.__name__}({repr(self.mcnp_str())})"
 
     def mcnp_str(self):
         """
@@ -191,7 +191,7 @@ class Isotope:
         """
         return self.Z * 1000 + self.A
 
-    def __repr__(self):
+    def __str__(self):
         return f"ZAID={self.ZAID}, Z={self.Z}, A={self.A}, element={self.element}, library={self.library}"
 
     def __hash__(self):
--- a/montepy/data_inputs/isotope.py
+++ b/montepy/data_inputs/isotope.py
@@ -181,7 +181,8 @@ class Isotope:
         return f"{self.ZAID}.{self.library}"
 
     def nuclide_str(self):
-        return f"{self.element.symbol}-{self.A}.{self._library}"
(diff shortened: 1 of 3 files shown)


tests/test_material.py
```diff
@@ -98,16 +98,17 @@ def test_material_str(self):
...
         material = Material(input_card)
-        answers = """MATERIAL: 20 fractions: atom
- H-1   (80c) 0.5
- O-16  (80c) 0.4
-Pu-239 (80c) 0.1
+        answers = """\
+MATERIAL: 20 fractions: atom
+ZAID=1001, Z=1, A=1, element=hydrogen, library=80c 0.5
+ZAID=8016, Z=8, A=16, element=oxygen, library=80c 0.4
+ZAID=94239, Z=94, A=239, element=plutonium, library=80c 0.1
```
MicahGale, line 105: I think keeping this is makes a worse representation. I think the old `repr` should just be removed and either the old `str` be kept, or a simpler form be used.

montepy/data_inputs/isotope.py
```diff
@@ -166,8 +166,8 @@ def library(self, library):
             raise TypeError("library must be a string")
         self._library = library
 
-    def __str__(self):
-        return f"{self.element.symbol:>2}-{self.A:<3} ({self._library})"
+    def __repr__(self):
+        return f"{self.__class__.__name__}({repr(self.mcnp_str())})"
```
MicahGale, line 170: I'm not sure if I like using ZAIDs for `repr`. I would prefer `Isotope('Pu239.80c')`, though this is invalid MCNP.

MicahGale (changes requested):

tests/test_material.py
tjlaboss: This is the `__str__`. I'll change it to the old `__repr__` and update the material test accordingly.

MicahGale: This still feels very verbose.

montepy/data_inputs/isotope.py
tjlaboss: I like that better, too. I'll use the hyph-en-dash variant, because that is valid MCNP 6.3.1

tjlaboss: [0cdc143](https://github.com/idaholab/MontePy/pull/474/commits/0cdc14374a9df49dcd05307da3fbdef289354154)

tests/test_material.py
tjlaboss: See the new `Isotope.nuclide_str()` and let me know what you think

montepy/data_inputs/isotope.py
```diff
@@ -180,6 +180,9 @@ def mcnp_str(self):
         """
         return f"{self.ZAID}.{self.library}"
 
+    def nuclide_str(self):
+        return f"{self.element.symbol}-{self.A}.{self._library}"
+
```
MicahGale, line 185: I like this. Except library can be None. or Empty. I'd add an else clause that drops the `.`.

tjlaboss: Good point. Just accounted for that.

tjlaboss:
Ready for Round II review

tests/test_material.py
tjlaboss: kept old `__str__`

tests/test_material.py
```diff
@@ -223,14 +224,12 @@ def test_isotope_library_setter(self):
...
-            repr(isotope), "ZAID=1001, Z=1, A=1, element=hydrogen, library=80c"
-        )
+        assert isotope.mcnp_str() == "1001.80c"
+        assert repr(isotope) == "Isotope('H-1.80c')"
+        assert str(isotope) == " H-1   (80c)"
         isotope = Isotope("94239.80c")
-        self.assertEqual(isotope.mcnp_str(), "94239.80c")
-        self.assertEqual(str(isotope), "Pu-239 (80c)")
+        assert isotope.mcnp_str() == "94239.80c"
+        assert repr(isotope) == "Isotope('Pu-239.80c')"
```
MicahGale, line 232: Also test `nuclide_str`.

tjlaboss: Tested `nuclide_str()` and the case where there is no library set

tjlaboss:
This PR also adds a test for Am-242m1 (`95642`), which is currently presented as `"Am-242"`.

tests/test_material.py
```diff
@@ -245,14 +246,18 @@ def test_isotope_library_setter(self):
...
         isotope = Isotope("94239.80c")
-        self.assertEqual(isotope.mcnp_str(), "94239.80c")
-        self.assertEqual(str(isotope), "Pu-239 (80c)")
+        assert isotope.nuclide_str() == "Pu-239.80c"
+        assert isotope.mcnp_str() == "94239.80c"
+        assert repr(isotope) == "Isotope('Pu-239.80c')"
+        isotope = Isotope("95642")
+        assert isotope.nuclide_str() == "Am-242"
+        assert isotope.mcnp_str() == "95642"
+        assert repr(isotope) == "Isotope('Am-242')"
```
MicahGale, line 260: This feels misleading, but I think that's a different issue to open.

tjlaboss: I agree. This test will be modified to fail in a separate branch.

PR #491, opened 2024-08-16
Bug fix for parentheses getting dropped from cell definitions. 

### LINKED ISSUE #490: MontePy deletes parentheses from cell regions
**Describe the bug**

MontePy discards the parentheses from region definitions when writing cell inputs. This appears to take place even when the parentheses are significant.

It is important to note that it parses the region correctly upon reading.

**To Reproduce**

``` python
import montepy
deck1 = montepy.read_input("testi.i")
print("Original Cell 3 Region")
print(deck1.cells[3].geometry)
deck1.write_to_file("testo.i", overwrite=True)
deck2 = montepy.read_input("testo.i")
print("Incorrect Cell 3 Region")
print(deck2.cells[3].geometry)
```

**Output**

```
Original Cell 3 Region
(((-1010*+1001)*+13)*(((-1002:-1003):-1004):-1005))
Incorrect Cell 3 Region
((((((-1010*+1001)*+13)*-1002):-1003):-1004):-1005)
```

##### _testi.i_

```
MCNP Test Model
C Cells with parentheses
1 3   1  (-1001 +11 +13):(-1001 -11 -13)   imp:n=2
2 2   1 ((-1001 +11 -13):(-1001 -11 +13))  imp:n=2
3 1 -10 (   -1010                                   $ inside big sphere
            +1001                                   $ outside origin sphere
            +13                                     $ above Z-axis
           (-1002:-1003:-1004:-1005)    )           $ inside the 4 spheres
           imp:n=4
4 1 -10 -1010 +1001 -13 (-1002:-1003:-1004:-1005)   $ like 3 but below the z-plane
           imp:n=1
5     0 -1010 1001 1002 1003 1004 1005     imp:n=2  $ outside the spheres
C Graveyard
99    0 +1010     imp:n=0

C Surfaces
C
C Planes
11  PX  0
12  PY  0
13  PZ  0
C
C Spheres
1001 SO      0.6
1002 SX -1.1 0.1
1003 SX +1.1 0.1
1004 SY -1.1 0.1
1005 SY +1.1 0.1
1010 SO 3

C data
C materials
C UO2 5 atpt enriched
m1        92235.80c           5
          92238.80c          95
C Helium
m2         2004.80c           1
C Helium-3
m3         2003.80c           1
C execution
ksrc -1.1 0 0.01
kcode 100000 1.000 50 1050
mode n
vol NO 2J 1 1.5 J
```

##### Cell block of _testo.i_

```
MCNP Test Model
C Cells with parentheses
1 3   1  -1001 +11 +13:-1001 -11 -13 imp:n=2 
2 2   1 -1001 +11 -13:-1001 -11 +13 imp:n=2 
3 1 -10 -1010    
            +1001
            +13
           -1002:-1003:-1004:-1005    imp:n=4
4 1 -10 -1010 +1001 -13 -1002:-1003:-1004:-1005   $ like 3 but below the z-plane
           imp:n=1  
5     0 -1010 1001 1002 1003 1004 1005     imp:n=2  $ outside the spheres
C Graveyard
99    0 +1010     imp:n=0
```

**Version**

 - Version 0.2 <= 0.3.3

**Additional context**

Some (but not all) in-line comments are deleted, too.

MicahGale:
# Description

This fixes a bug where parentheses disappear from the geometry definition of a cell. This was because `surfaces.HalfSpace` essentially had no way to store that information, and when it was time to export the geometry would override the old parentheses. 

This was solved in a few ways.

1. Introduce `Operation.GROUP` to replace `Operation._SHIFT` in this context. This was because this will be more visible to users and `GROUP` is clearer. Also `_SHIFT` is used for certain leaf nodes in complex geometry, and lead to confusion.
2. Adding an unbalanced tree level to `HalfSpace` to represent `GROUP`
3. Updating `HalfSapce` to ensure that `GROUP` is properly handled on export.
4. Made a method to detect when a group needs to be added for scratch geometries. This is done whenever a union node is below an intersection node. This is the proper order of operations (see #490). 

To sanity check this logic:

```
    *
   / \
  +   3
 /\
1  2
```
Should be `(1 : 2) 3` not `1 : 2 3`

Fixes #490

# Checklist

- [x] Implement unit testing. 
- [x] Clear out all `TODO`
- [x] Look into disappearing comments with @tjlaboss

### DIFF
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -195,7 +195,7 @@ def test_the_dissapearing_parens():
                 if char in {"(", ")"}:
                     parens_count += 1
     fh = io.StringIO()
-    problem.write_to_file(fh)
+    problem.write_problem(fh)
     new_parens_count = 0
     fh.seek(0)
     for line in fh:
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -196,9 +196,11 @@ def test_the_dissapearing_parens():
     problem.write_problem(fh)
     new_parens_count = 0
     fh.seek(0)
-    for line in fh:
-        print(line.rstrip())
-        new_parens_count += line.count("(") + line.count(")")
+    with open(in_file, "r") as gold_fh:
+        for line, gold_line in zip(fh, gold_fh):
+            print(line.rstrip())
+            new_parens_count += line.count("(") + line.count(")")
+            assert line.rstrip() == gold_line.rstrip()
     assert new_parens_count == parens_count
 
 
@@ -213,7 +215,6 @@ def test_universe_after_comment():
     try:
         out_file = "universe_after_comment"
         problem.write_to_file(out_file)
-        # TODO check that the Universe does show up
         found = False
         with open(out_file, "r") as fh:
             for line in fh:
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -187,3 +187,29 @@ def test_complex_cell_parsing(lines):
         montepy.input_parser.block_type.BlockType.CELL,
     )
     montepy.Cell(input)
+
+def test_the_dissapearing_parens():
+    in_file = os.path.join("tests", "inputs", "test_paren_groups.imcnp")
+    problem = montepy.read_input(in_file)
+    parens_count = 0
+    with open(in_file, "r") as fh:
+        for line in fh:
+            for char in line:
+                if char in {"(", ")"}:
+                    parens_count += 1
+    try:
+        out_file = "test_parens"
+        problem.write_to_file(out_file)
+        new_parens_count = 0
+        with open(out_file, "r") as fh:
+            for line in fh:
+                print(line.rstrip())
+                for char in line:
+                    if char in {"(", ")"}:
+                        new_parens_count += 1
+        assert new_parens_count == parens_count
+    finally:
+        try:
+            os.remove(out_file)
+        except FileNotFoundError:
+            pass
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -184,6 +184,7 @@ def test_complex_cell_parsing(lines):
     )
     montepy.Cell(input)
 
+
 def test_the_dissapearing_parens():
     in_file = os.path.join("tests", "inputs", "test_paren_groups.imcnp")
     problem = montepy.read_input(in_file)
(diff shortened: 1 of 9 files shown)


MicahGale:
Should we test writing all files out to file at least once and see what changes?

tests/test_edge_cases.py
```diff
@@ -187,3 +187,29 @@ def test_complex_cell_parsing(lines):
...
+        out_file = "test_parens"
+        problem.write_to_file(out_file)
+        new_parens_count = 0
+        with open(out_file, "r") as fh:
+            for line in fh:
+                print(line.rstrip())
+                for char in line:
+                    if char in {"(", ")"}:
+                        new_parens_count += 1
+        assert new_parens_count == parens_count
```
MicahGale, line 210: Is this a janky way to test?

tests/test_edge_cases.py
```diff
@@ -187,3 +187,29 @@ def test_complex_cell_parsing(lines):
...
+    finally:
+        try:
+            os.remove(out_file)
+        except FileNotFoundError:
+            pass
```
MicahGale, line 215: I'd like thoughts on how to fixture this try finally block. I write it a ton.

tjlaboss: [I have a better idea.](https://github.com/idaholab/MontePy/issues/492)

tests/test_edge_cases.py
tjlaboss: [`str.count`](https://docs.python.org/3/library/stdtypes.html#str.count)

tjlaboss:
> Should we test writing all files out to file at least once and see what changes?

Eventually, we should test writing out each type of input at least once. This can be done piecemeal, with `file read=...` coming last.

tests/test_edge_cases.py
MicahGale: Yeah before writing this I was pondering if `IO.StringIO` would work.

MicahGale:
> > Should we test writing all files out to file at least once and see what changes?
> 
> Eventually, we should test writing out each type of input at least once. This can be done piecemeal, with `file read=...` coming last.

I think this should wait until `IO.StringIO` is supported.

We already bulk read all valid test inputs through the checker. There's already good machinery for finding all valid input files. I think doing this enmass would be a good idea. We just run the risk of discovering a lot of bugs all at once that might is out of scope for this work I think.

tests/test_edge_cases.py
tjlaboss: It will 🤠

tjlaboss (approved):
NICE

PR #493, opened 2024-08-16
Write to stream

### LINKED ISSUE #492: Implement MCNP_Problem.write(file_handle)

In addition to `MCNP_Problem.write_to_file(new_problem)`, implement `MCNP_Problem.write_to_stream(file_stream)`. Then, users can write to any open file stream, including `io.StringIO`.

```python
class MCNP_Problem:
    def write_to_file(self, str: file_path, bool: overwrite=False):
        """
        Check for existence
        Open the file handle
        Call self.write()`
        Close the file handle

        Also, rename kwarg from "new_problem" to "file_path" or something that
        does not imply that it is another ``MCNP_Problem`` instance.
        """
       . . .
```

and

```python
class MCNP_Problem:
    def write_to_stream(io.TextIOBase: file_handle):
        """
        Write to an open stream.

        Can allow to fail naturally, or do either of the following checks:
        >>> isinstance(file_handle, io.TextIOBase)
        >>> hasattr(file_handle, "write") and callable(getattr(file_handle, "write"))
        """
        . . .
```

The exact naming is up for discussion. I like `fpath` instead of `file_path` and `fhandle` or `fh` instead of `file_handle`.

Maybe `MCNP_Problem.write_to_stream()` should return something, such as the difference of `f.tell()` before and after writing in case the user wants to rewind the file (similar to a stream's `.write()`).

Another alternative is that `MCNP_Problem` could function more like a stream itself, with its own `.read()` and `.write()` methods. This seems unnecessary.

**Additional context**
The `.write_to_stream()` method would be highly useful in the MontePy integration tests.

tjlaboss:
# Description

* Implement `MCNP_Problem._write_to_stream()` 
* Implement `MCNP_Problem.write_problem()`, which handles both file paths and IO streams.
* Mark `MCNP_Problem.write_to_file()` with a deprecation warning.
* Update the integration and edge case tests accordingly.
* Include `montepy.errors` in top-level namespace
* Convert _test_integration.py_ from unittest to pytest, fixing some minor errors along the way.
* Update documentation accordingly.

Fixes #492

# Checklist

- [x] Either fully implement or deprecate `MCNP_InputFile`
- [x] Determine what `MCNP_Problem.write_*` methods should return
- [x] Settle on names for affected methods and parameters
- [x] I have made corresponding changes to the documentation
- [x] I have updated the changelog

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -1,12 +1,16 @@
 MontePy Changelog
 =================
 
-#Next Version#
---------------
-
+0.4.0
+-----
 **Bug Fixes**
 
 * Fixed bug that didn't show metastable states for pretty printing and isotope. Also handled the case that Am-241 metstable states break convention (:issue:`486`).
+
+**Features Added**
+
+* Write problems to either file paths or streams (file handles) with MCNP_Problem.write_problem() (:issue:`492`).
+
 
 0.3.3
 --------------
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -1,15 +1,15 @@
 MontePy Changelog
 =================
 
-0.4.0
------
-**Bug Fixes**
-
-* Fixed bug that didn't show metastable states for pretty printing and isotope. Also handled the case that Am-241 metstable states break convention (:issue:`486`).
-
+#Next Version#
+--------------
 **Features Added**
 
 * Write problems to either file paths or streams (file handles) with MCNP_Problem.write_problem() (:issue:`492`).
+
+**Bug Fixes**
+
+* Fixed bug that didn't show metastable states for pretty printing and isotope. Also handled the case that Am-241 metstable states break convention (:issue:`486`).
 
 
 0.3.3
--- a/montepy/input_parser/input_file.py
+++ b/montepy/input_parser/input_file.py
@@ -31,12 +31,12 @@ class MCNP_InputFile:
         self._overwrite = overwrite
         self._mode = None
         self._fh = None
-    
+
     @classmethod
     def from_open_stream(cls, fh):
         """
         Create an MCNP Input File from an open, writable stream
-        
+
         :param fh: An open and writable object, such as a file handle.
         :type fh: io.TextIOBase
         """
--- a/montepy/input_parser/input_file.py
+++ b/montepy/input_parser/input_file.py
@@ -31,6 +31,19 @@ class MCNP_InputFile:
         self._overwrite = overwrite
         self._mode = None
         self._fh = None
+    
+    @classmethod
+    def from_open_stream(cls, fh):
+        """
+        Create an MCNP Input File from an open, writable stream
+        
+        :param fh: An open and writable object, such as a file handle.
+        :type fh: io.TextIOBase
+        """
+        name = getattr(fh, "name", fh.__class__.__name__)
+        inpfile = cls(path=name)
+        inpfile._fh = fh
+        return inpfile
 
     @make_prop_pointer("_path")
     def path(self):
--- a/montepy/mcnp_problem.py
+++ b/montepy/mcnp_problem.py
@@ -390,11 +390,11 @@ class MCNP_Problem:
         self._materials = Materials(materials)
         self._transforms = Transforms(transforms)
         self._data_inputs = sorted(set(self._data_inputs + materials + transforms))
-    
+
     def write_problem(self, destination, overwrite=False):
         """
         Write the problem to a file or writeable object.
-        
+
         :param destination: File path or writable object
         :type destination: object
         :param overwrite: Whether to overwrite 'destination' if it is an existing file
@@ -405,7 +405,9 @@ class MCNP_Problem:
             return self.write_to_stream(destination)
         if isinstance(destination, (str, os.PathLike)):
             return self.write_to_file(destination, overwrite)
-        raise TypeError(f"destination f{destination} is not a file path or writable object")
+        raise TypeError(
+            f"destination f{destination} is not a file path or writable object"
+        )
 
     def write_to_file(self, new_problem, overwrite=False):
         """
@@ -426,14 +428,14 @@ class MCNP_Problem:
         with new_file.open("w") as fh:
             self.write_to_stream(fh)
         # Consider returning result of write_to_stream()
-    
+
     def write_to_stream(self, fh):
         """
         Writes the problem to a writeable stream.
-        
+
         TODO: Expand MCNP_InputFile or adjust this method
               to fully suppport generic file handles.
-        
+
         :param fh: Writable object
         :type fh: {MCNP_InputFile, io.TextIOBase}
         """
--- a/montepy/mcnp_problem.py
+++ b/montepy/mcnp_problem.py
@@ -406,9 +406,17 @@ class MCNP_Problem:
         :raises IsADirectoryError: if the path given is actually a directory.
         """
         new_file = MCNP_InputFile(new_problem, overwrite=overwrite)
-        with new_file.open("w") as fh, warnings.catch_warnings(
-            record=True
-        ) as warning_catch:
+        with new_file.open("w") as fh:
+            self.write_to_stream(fh)
+    
+    def write_to_stream(self, fh):
+        """
+        Writes the problem to a writeable stream.
+        
+        :param fh: Writable object
+        :type fh: {MCNP_InputFile, io.TextIOBase}
+        """
+        with warnings.catch_warnings(record=True) as warning_catch:
             objects_list = []
             if self.message:
                 objects_list.append(([self.message], False))
@@ -442,6 +450,7 @@ class MCNP_Problem:
 
             fh.write("\n")
         self._handle_warnings(warning_catch)
+        # Todo: Consider implementing and calling MCNP_InputFile.tell()
 
     def _handle_warnings(self, warning_queue):
         class WarningLevels(Enum):
--- a/montepy/mcnp_problem.py
+++ b/montepy/mcnp_problem.py
@@ -396,7 +396,7 @@ class MCNP_Problem:
         Write the problem to a file or writeable object.
 
         :param destination: File path or writable object
-        :type destination: {io.TextIOBase, str, os.PathLike}
+        :type destination: io.TextIOBase, str, os.PathLike
         :param overwrite: Whether to overwrite 'destination' if it is an existing file
         :type overwrite: bool
         """
@@ -423,7 +423,7 @@ class MCNP_Problem:
             The overwrite parameter was added.
 
         :param file_path: the file path to write this problem to
-        :type file_path: {str, os.PathLike}
+        :type file_path: str, os.PathLike
         :param overwrite: Whether to overwrite the file at 'new_problem' if it exists
         :type overwrite: bool
         :raises IllegalState: if an object in the problem has not been fully initialized.
--- a/montepy/mcnp_problem.py
+++ b/montepy/mcnp_problem.py
@@ -415,6 +415,9 @@ class MCNP_Problem:
     def write_to_file(self, file_path, overwrite=False):
         """
         Writes the problem to a file.
+        
+        .. versionchanged:: 0.4.0
+            Deprecated in favor of self.write_problem()
 
         .. versionchanged:: 0.3.0
             The overwrite parameter was added.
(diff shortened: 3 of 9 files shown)


montepy/mcnp_problem.py
```diff
@@ -390,25 +391,56 @@ def add_cell_children_to_problem(self):
         self._transforms = Transforms(transforms)
         self._data_inputs = sorted(set(self._data_inputs + materials + transforms))
 
-    def write_to_file(self, new_problem, overwrite=False):
+    def write_problem(self, destination, overwrite=False):
+        """
+        Write the problem to a file or writeable object.
+
```
MicahGale, line 397: I'm not sure if this is the most pythonic name.

MicahGale: * Openpyxl: `save()` not a fan. This is very MS office centric
* pandas: `to_*()` e.g., `to_excel` `to_json`... I think this makes sense for something so filetype agnostic
* numpy: `tofile()`, `dump`
* openmc: `export_to_xml`
* json/pickle: `dump`

So nvm, there isn't a really "pythonic" way to do it. I think MontePy just needs to be consistent

So I propose:

1. `write` means to write out the state in a way that MCNP understands
2. `dump` means to serialize to a file (JSON or pickle)

`write` usually takes a string as an argument, so I think `write_problem` is a good name.

montepy/mcnp_problem.py
```diff
@@ -390,25 +391,56 @@ def add_cell_children_to_problem(self):
...
         """
         Writes the problem to a file.
 
         .. versionchanged:: 0.3.0
             The overwrite parameter was added.
 
-        :param new_problem: the file name to write this problem to
-        :type new_problem: str
+        :param file_path: the file path to write this problem to
+        :type file_path: str
```
MicahGale, line 420: Probably should add `os.pathlike`

montepy/mcnp_problem.py
```diff
@@ -390,25 +391,56 @@ def add_cell_children_to_problem(self):
...
         :raises FileExistsError: if a file already exists with the same path.
         :raises IsADirectoryError: if the path given is actually a directory.
         """
-        new_file = MCNP_InputFile(new_problem, overwrite=overwrite)
-        with new_file.open("w") as fh, warnings.catch_warnings(
-            record=True
-        ) as warning_catch:
+        new_file = MCNP_InputFile(file_path, overwrite=overwrite)
+        with new_file.open("w") as fh:
+            self.write_to_stream(fh)
```
MicahGale, line 429: I was about to freak about losing the warning_catch, but now I see what you did. I like it.

montepy/mcnp_problem.py
```diff
@@ -390,25 +391,56 @@ def add_cell_children_to_problem(self):
...
+        # Consider returning result of write_to_stream()
+
+    def write_to_stream(self, fh):
+        """
+        Writes the problem to a writeable stream.
+
+        TODO: Expand MCNP_InputFile or adjust this method
+              to fully suppport generic file handles.
```
MicahGale, line 437: I think that's the ultimate solution here. How it is getting the input file lineno is just `\n` counting, which we can totally do with a stream. 

As for `tell` I don't think it's necessary. `tell` has no correlation to line number (or even character count sometimes, and is only useful for `seek`. I don't think we should implement `seek`.

MicahGale (commented):
Ok I like the interface for the users.

After reading everything I think the approach should be:

1. implement stream support for `MCNP_InputFile`
2. make `write_to_*` prepare an `MCNP_InputFile`
3. call a `_private` method with the InputFile in the right state that then handles the warnings.

montepy/mcnp_problem.py
tjlaboss: That was my exact train of thought.

tjlaboss:
The reported test failures on a56f2943522ba5a0df1534ba0287ddf22398f88c are from the deprecation warnings. All the unit tests actually pass.

I've implemented a new class method `MCNP_InputFile.from_open_stream()` that bypasses `MCNP_InputFile.open()`. Take a look and see what you think. If you like the functionality, let's think about the interface. In my opinion, `.__init__()` should actually be based on a stream. Then `.open()` could be converted to a `@classmethod` that creates the file handle and instantiates a new object.

MicahGale:
TODO: update the test from #491 at the very least.

MicahGale:
> The reported test failures on [a56f294](https://github.com/idaholab/MontePy/commit/a56f2943522ba5a0df1534ba0287ddf22398f88c) are from the deprecation warnings. All the unit tests actually pass.
> 
> I've implemented a new class method `MCNP_InputFile.from_open_stream()` that bypasses `MCNP_InputFile.open()`. Take a look and see what you think. If you like the functionality, let's think about the interface. In my opinion, `.__init__()` should actually be based on a stream. Then `.open()` could be converted to a `@classmethod` that creates the file handle and instantiates a new object.

Hmm interesting idea. I'm a bit hesitant this second. Look into how `read_input` works with this class first.

doc/source/changelog.rst
```diff
@@ -1,13 +1,17 @@
 MontePy Changelog
 =================
 
-#Next Version#
---------------
-
+0.4.0
+-----
```
MicahGale, line 5: Not the time or place to change this.

doc/source/changelog.rst
```diff
@@ -1,13 +1,17 @@
...
 **Bug Fixes**
 
 * Fixed bug that didn't show metastable states for pretty printing and isotope. Also handled the case that Am-241 metstable states break convention (:issue:`486`).
 
+**Features Added**
+
+* Write problems to either file paths or streams (file handles) with MCNP_Problem.write_problem() (:issue:`492`).
+
+
```
MicahGale, line 14: Features before bugs.

montepy/input_parser/input_file.py
```diff
@@ -32,6 +32,19 @@ def __init__(self, path, parent_file=None, overwrite=False):
...
 
+    @classmethod
+    def from_open_stream(cls, fh):
+        """
+        Create an MCNP Input File from an open, writable stream
+
+        :param fh: An open and writable object, such as a file handle.
+        :type fh: io.TextIOBase
+        """
+        name = getattr(fh, "name", fh.__class__.__name__)
```
MicahGale, line 43: Would `type(fh).__name__` be more pythonic?

montepy/mcnp_problem.py
```diff
@@ -390,25 +391,58 @@ def add_cell_children_to_problem(self):
         self._transforms = Transforms(transforms)
         self._data_inputs = sorted(set(self._data_inputs + materials + transforms))
 
-    def write_to_file(self, new_problem, overwrite=False):
+    def write_problem(self, destination, overwrite=False):
+        """
+        Write the problem to a file or writeable object.
+
+        :param destination: File path or writable object
+        :type destination: {io.TextIOBase, str, os.PathLike}
```
MicahGale, line 399: Is this proper sphinx?

montepy/mcnp_problem.py
```diff
@@ -390,25 +391,58 @@ def add_cell_children_to_problem(self):
...
+                f"destination f{destination} is not a file path or writable object"
+            )
+
+    def write_to_file(self, file_path, overwrite=False):
         """
         Writes the problem to a file.
 
+        .. versionchanged:: 0.4.0
+            Deprecated in favor of self.write_problem()
+
```
MicahGale, line 421: Should be a `.. deprecated::` then.

montepy/mcnp_problem.py
```diff
@@ -390,25 +391,58 @@ def add_cell_children_to_problem(self):
...
         :raises FileExistsError: if a file already exists with the same path.
         :raises IsADirectoryError: if the path given is actually a directory.
         """
-        new_file = MCNP_InputFile(new_problem, overwrite=overwrite)
-        with new_file.open("w") as fh, warnings.catch_warnings(
-            record=True
-        ) as warning_catch:
+        cn = self.__class__.__name__
+        msg = f"{cn}.write_to_file() is deprecated. Use {cn}.write_problem()"
+        warnings.warn(msg, DeprecationWarning)
```
MicahGale, line 435: I'm not sure if we need to raise deprecation warnings yet.

tests/test_edge_cases.py
```diff
@@ -41,13 +41,9 @@ def test_orphaning_mt(self):
             montepy.input_parser.block_type.BlockType.DATA,
         )
         problem.data_inputs.append(montepy.data_inputs.data_parser.parse_data(card))
-        try:
+        with io.StringIO() as stream:
             with self.assertRaises(MalformedInputError):
```
MicahGale, line 45: `pytest.raises`

tests/test_integration.py
MicahGale, line 1: Did you add any new tests? TL;DR

MicahGale (changes requested):

MicahGale:
Coverage has gone down slightly.

doc/source/changelog.rst
tjlaboss: Should I separate them into "#Next Minor Version#" and "#Next Patch#"?

montepy/input_parser/input_file.py
tjlaboss: I generally use `.__class__.__name__` for printing, to respect an overwrite of `.__class__` if that should occur, but have no strong feelings on the matter.

montepy/mcnp_problem.py
tjlaboss: `PendingDeprecationWarning`, or no warning at all?

tests/test_integration.py
tjlaboss: The existing tests have been moved to use `write_problem()` in all but one case, which continues to test `write_to_file()` with `pytest.warns(DeprecationWarning)`.

`test_importance_rewrite()` tests `write_problem()` with two real files. The rest test `write_problem()` with a stream, specifically a `StringIO`.

tests/test_integration.py
```diff
@@ -1,1048 +1,1069 @@
...
-        surf.transform.number = 5
-        self.assertEqual(int(surf.format_for_mcnp_input((6, 2, 0))[0].split()[1]), 5)
-        # test changing surface constants
-        new_prob = copy.deepcopy(problem)
-        surf = new_prob.surfaces[4]
-        surf.location = 2.5
-        self.assertEqual(
-            float(surf.format_for_mcnp_input((6, 2, 0))[0].split()[-1]), 2.5
-        )
+@pytest.fixture
```
MicahGale, line 26: You should make this have a scope so it's ran once. Not being able to cache these problems really slowed down the test suite for a while. This is what I came up with recently: https://github.com/idaholab/MontePy/pull/488/files#diff-4071f2914f1a922463edf531fe1ccb42d1f0534ecb7f22b7812d955c840f1f46R295-R304

doc/source/changelog.rst
MicahGale: No the plan is to do one big release soon

MicahGale: #Next Version# is used by `grep`

montepy/input_parser/input_file.py
MicahGale: Ok that's good enough for me.

tests/test_integration.py
tjlaboss: I just added a new test that ensures that `write_problem()` catches an unwritable object.

montepy/mcnp_problem.py
MicahGale: pending is an option? I'm trying to think: what would @bascandr say?

I honestly don't see any reason to truly deprecate it and remove it. I think no warning. I'm fine with users not making changes at this point.

tests/test_integration.py
MicahGale: I think you'll need to write a few more tests to ensure 100% coverage of the changes.

montepy/mcnp_problem.py
tjlaboss: undid deprecation

tests/test_integration.py
tjlaboss: All new code is now covered

tests/test_integration.py
tjlaboss: Set the scope with [5460cb2](https://github.com/idaholab/MontePy/pull/493/commits/5460cb22d0ba8fcc63a88c5e667a223327196671)

After all the tests have been converted to pytest, we should consider making a global fixtures module. `test.imcnp` gets used a lot, for instance.

tjlaboss:
Re-ady for re-review.

MicahGale:
> Re-ady for re-review.

image
There's a button for that.

tjlaboss:
the button doesn't come with a quirky message tho

MicahGale:
Touche.

MicahGale (approved):
LGTM. 

In the feature I can't decide if most test cases should move to `IO.StringIO`.

PR #507, opened 2024-08-25
Mat redesign

### LINKED ISSUE #63: Remove duplicate objects without relying on Hashing
Right now duplicates are removed by using hashes and the set operator. However these objects are mutable and these values changing can lead to unpredictable results. See: https://eng.lyft.com/hashing-and-equality-in-python-2ea8c738fb9d.

This affects most `NumberdObjectCollection` children objects.

### LINKED ISSUE #95: Allow querying materials by elements
I think that the `Materials` collection should have a filtering generator similar to how `Surfaces` has one for surface type (e.g, `surfaces.pz`). 

I envisage that you should be able to filter by containing elements. e.g., `materials.u` or `materials.uranium` would return all materials that contain 92***.**c, etc. I can't decide if it should be done by element atomic symbol or by element name. Name is easier to read and doesn't require memorizing atomic symbols. Symbols would be shorter and avoids the whole controversy of UK-English spelling aluminum wrong.

### LINKED ISSUE #112: Separate nuclear data from the concept of an Isotope
Right now the `Isotope` object both represents an isotope, but also a specific set of nuclear data representing that isotope (e.g., the `1001.80c` ACE data). This is complicating things and muddying the water. Isotopes need to be sovereign and independent of the nuclear data (but probably really should represent an isomer). Then a nuclear data object (name TBD) can wrap that isotope inside a specific nuclear data library at a specific temperature. 

This complicated my life because when adding metastable isomers to the `Isotope` object (#70) I wanted to add a function: `def get_base_isotope(self): -> Isotope` however with the nuclear data library information this function can have weird consequences and therefore has been tabled for the time being.

### LINKED ISSUE #138: Make NumberedObjectCollection act like a set
This is very similar to #61. I am partially of the mind of either implementing both or only implement set behavior and not dict behavior.

This came out of this discussion (copied below): https://hpcgitlab.hpc.inl.gov/experiment_analysis/mcnpy/-/issues/54#note_480927h.

To summarize: implement set-like arithmetic and binary logic. This will leads to complications with multiple problems though and #22, and #83 will be involved in the solution.

Labossiere-Hickman, Travis James
@tjlaboss
> I do like the option of having a NumberedObjectCollection behave as a set. Since the collections can come from different models, though, the standard Python set logic might not be appropriate for identically-numbered but unequal objects. Three solutions:

> Only allow set operations for NumberedObjectCollection instances linked to the same problem.
> Allow them, but warn for every conflict.
> Don't implement set operations, and just have the user manually do set(num_obj_col) if they want.

Gale, Micah David
@MicahGale

> Just going to mention #61 here.

> But this is very much an issue that I have been grappling for awhile, that lead to all those weird scripts that can't be upgraded that I showed you today @bascandr. This needs to ultimately get resolved by #22.

> My overall view is that we should allow different problems. I think we need to be able to spot ostensibly identical objects to merge (for instance two files with common heritage). If they are the same they should merge. This then deals with problems of pointers though. Maybe we need to implement a hidden object state that just engages passthrough mode, probably knock that sort of thing out along with #83. But if there is a number collision of dissimilar objects I think the RHS (to be consistent) should be renumbered with a warning. Though this could cause issues if users didn't intend that. Maybe we need a system to rollback changes? Like a database? That would be actually really nice:

```python
try: 
   problem.start_transaction()
   ... do error prone thing...
   problem.commit()
except as e:
   problem.roll_back()
   raise e
```

> Realistically this would be a with statement most likely.

> Suffice it to say I think we enough discussion to open a few new issues.

### LINKED ISSUE #144: Material strings can become unruly.
When you print a material a list of all its elements are included. For depleted fuel this can become ... very unruly. Should implement a numpy style snip that cuts out a lot of the middle bits.

### LINKED ISSUE #182: Support poorly formatted Material definitions
LANL apparently doesn't know how to write a syntax specification. From LA-UR-17-29981 Section 3.3.2.1 a material specification shall take the form:

> Mm zaid1 fraction1 zaid2 fraction2 ... [KEYWORD=value(s) ...]

Which makes sense. 

Until you notice

> | Input Parameter | Description |
> |-----------------|-------------|
> | Keyword *       | Value       |

And finally notice the asterisk:

> \* M card keywords may appear anywhere among the zaid-fraction pairs, but must not separate a pair.

Yep The syntax specification from before is a lie. The key-value pairs can appear anywhere they like!

Why did the developers do this? I don't know.

Is this a context-free grammar? Probably not.

However, it is officially (in the worst documented ways possible) supported, so it needs to be supported in MOntePy.

### LINKED ISSUE #369: Add support for particle specific libraries and default libraries.
This is very closely related to #21, and should be implemented at the same time. 

Features to support:

1. `M0` according to Note 1 of section 3.3.2.1 of the 6.2 manual an `M0` material can be specified to specify the default libraries to use.
2. Particle specific libraries. According to Table 3-32 you can specify `plib`, `nlib`, etc. I think this should be specified similar to importances: i.e., `mat.library.neutron`
3. Default to first `XSDIR` entry. This is described in the intro to Section 3.3.2.1. This feature would require #113. The idea being if no library is specified, no `M0`, and an `XSDIR` is provided users can see what the default library will be. This should be implemented with caution to avoid misleading users about what is going on here.

### LINKED ISSUE #383: Moving to true Continuous Deployment: Creating an Alpha release channel.
### Discussed in https://github.com/idaholab/MontePy/discussions/371

It would be helpful to sometimes have an official alpha release for user testing. 

This would involve implementing the same CD process used for Master right now except the version would be marked as alpha e.g., `0.3.0a1`.

### LINKED ISSUE #504: Duplicate Nuclides in Material definition will lose information
**Describe the bug**

Recent discussions with @tjlaboss and @gonuke has brought to my attention that anecdotally material definitions might (and have good reason) repeat nuclides in the material definitions. For now I will say this is ambiguous.

MontePy was built with the assumption that this wasn't the case, and so cannot handle it properly. Material composition data are stored in a dictionary so the last instance of a nuclide will overwrite previous ones. Here a unique nuclide is a unique ZAID and a unique library. 

```
m1 1001.80c 0.5 1001.80c 0.5
```

Would export as:

```
m1 1001.80c 0.5
```

Whereas:

```
m1 1001.80c 0.5 1001.00c 0.5
```

Would be exported properly.

**Use Cases**

So there are two cases to cover:

1. Repeated nuclides with different libraries
1. Repeated nuclides with the same libraries

***Repeats with identical libraries***

This case is helpful for defining a mixed material where the two components contain the same nuclide. 
This would make the material definition more readable (by MCNP standards). For example boric acid, B(OH)_3, dissolved in water. 

***Repeats with different libraries***

This case can be used to mix two libraries with different temperatures as a means for temperature interpolation. 

**Research into the Manual**

The MCNP manual is vague on this topic. It says:

> Each material must be defined as a set of components with their corresponding material fraction.

This implies that nuclides can't be repeated; however not everyone is a mathematician so sets shouldn't be interpreted too strictly. Reading all of 3.3.2.1 (6.2 manual) there is no explicit guidance one way or another on this issue. So for now this should be treated as being ambiguous.

**Work around**

For the time being if a user has problems with this they can avoid it by ensuring their model has no duplicate nuclides with identical libraries by manually summing these repeats and replacing it with one entry.

**Next Steps**
1. Ask for clarification on this ambiguity in the documentation.

This came about from my work to imagine an improved interface for material_components on the branch `mat_redesign`, and discussing it in #475.

If we do need to support repeated nuclides I cannot think of a good non-backwards-compatibility-breaking fix that would preserve the interface of `Material.material_components`.

I think it might be time for a major release to fix this.

DIscussion of what that new interface will be go to #475.

**To Reproduce**

This would show up any time a file is read and then written:

``` python
import montepy
montepy.read_input("foo").write_problem("bar")
```

**Version**

 - All version <= 0.4.0

### LINKED ISSUE #505: Improve Isotope Instantiation

"I'm always frustrated when" I want to instantiate an `Isotope`, but have to remember that it lives at `montepy.data_inputs.isotope.Isotope`. It should be a top-level object.

Also annoying is that the representation uses the "Element-Z" convention, but instantiation has to use ZAID.

**Example**

```python
>>> montepy.data_inputs.isotope.Isotope("1001.00c")
Isotope('H-1.00c')

>>> Isotope('H-1.00c')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'Isotope' is not defined

>>> montepy.data_inputs.isotope.Isotope("H-1.00c")
Traceback (most recent call last):
  File "montepy/data_inputs/isotope.py", line 34, in __init__
    int(parts[0])
ValueError: invalid literal for int() with base 10: 'H-1'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "montepy/data_inputs/isotope.py", line 36, in __init__
    raise ValueError(f"ZAID: {ZAID} could not be parsed as a valid isotope")
ValueError: ZAID: H-1.00c could not be parsed as a valid isotope

```

 * Import the `Isotope` class into the top-level namespace, in _montepy/__init__.py_
 * Allow instantiation from, e.g., `montepy.Isotope("H-1.00c")`, `montepy.Isotope("H1.01c")`, etc.

**Additional context**
[Someone](https://github.com/tjlaboss) changed the `repr` in #473, but didn't bother to update `__init__` accordingly.

### LINKED ISSUE #508: Implement __slots__ to avoid user confusion
### Discussed in https://github.com/idaholab/MontePy/discussions/345

<div type='discussions-op-text'>

<sup>Originally posted by **MicahGale** August  2, 2023</sup>
TIL about slots in python: https://stackoverflow.com/questions/472000/usage-of-slots

Basically statically allocating python attributes. The benefit would be to guard against typos frustrating users. For instance `cell.materail = mat`.</div>

### LINKED ISSUE #511: Make `Material.is_atom_fraction` settable
``` python
Traceback (most recent call last):
  File "/home/mgale/dev/montepy/gen_big_model.py", line 12, in <module>
    gen_mega_mat()
  File "/home/mgale/dev/montepy/gen_big_model.py", line 6, in gen_mega_mat
    mat.is_atom_fraction = True
AttributeError: can't set attribute 'is_atom_fraction'
```

### LINKED ISSUE #512: Can't make a material from scratch and export it.
**Describe the bug**

**To Reproduce**

A short code snippet of what you have ran. Please change or remove any specific values or anything that can't be public. For example:

``` python
import montepy

def gen_mega_mat():
    mat = montepy.data_inputs.Material()
    mat.number = 1
    mat._is_atom_fraction = True
    for i in range(1,93):
        iso = montepy.data_inputs.isotope.Isotope(f"{int(i*1000 + i*2.1)}.80c")
        mat.material_components[iso] = montepy.data_inputs.material_component.MaterialComponent(iso, 0.01)
    print(mat.format_for_mcnp_input((6,3,0)))

gen_mega_mat()                           
```

**Error Message (if any)**

``` python
Press ENTER or type command to continue
  File "/home/mgale/dev/montepy/gen_big_model.py", line 10
    print(mat.format_for_mcnp_input((6,3,0))
         ^
SyntaxError: '(' was never closed

shell returned 1

Press ENTER or type command to continue
Traceback (most recent call last):
  File "/home/mgale/dev/montepy/gen_big_model.py", line 12, in <module>
    gen_mega_mat()
  File "/home/mgale/dev/montepy/gen_big_model.py", line 10, in gen_mega_mat
    print(mat.format_for_mcnp_input((6,3,0)))
  File "/home/mgale/dev/montepy/montepy/data_inputs/material.py", line 145, in format_for_mcnp_input
    lines = super().format_for_mcnp_input(mcnp_version)
  File "/home/mgale/dev/montepy/montepy/mcnp_object.py", line 130, in format_for_mcnp_input
    self._update_values()
  File "/home/mgale/dev/montepy/montepy/data_inputs/material.py", line 155, in _update_values
    self._tree.nodes["data"] = new_list
AttributeError: 'Material' object has no attribute '_tree'
```

**Version**

 - Version 0.4.0

### LINKED ISSUE #550: Implement singleton pattern for Immutable objects
There are a few immutable objects that are frequently used, namely: `Element`. There's no reason that these instances couldn't be shared so a "singleton-like" pattern might help save memory.

I confirmed this was note the case:

``` python
In [1]: import montepy

In [2]: el1= montepy.data_inputs.element.Element(92)

In [3]: el2= montepy.data_inputs.element.Element(92)

In [4]: hex(id(el1))
Out[4]: '0x7fa5c7d14410'

In [5]: hex(id(el2))
Out[5]: '0x7fa5f8b4f770'
```

### LINKED ISSUE #559: Make documentation rustier/ NumPy compliant
In [Rust](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components) there is an expectation that doc strings will include examples.

This could be a good idea to start including more and more in our doc strings.

Along with this we would need to implement [doctest](https://docs.pytest.org/en/7.1.x/how-to/doctest.html) in pytest. 

Is this a good paradigm?

MicahGale:
# Description

Design based on #475.

* [x] Fixes #504 
* [x] Fixes #505 
* [x] fixes #383
* [x] fixes #508 
* [x] fixes #144 
* [x] fixes #95 
* [x] fixes #369 
* [x] fixes #512 
* [x] fixes #511 
* [x] fixes #182
* [x] fixes #63
* [x] fixes #138
* [x] fixes #112
* [x] fixes #550 
* [x] Fixes #559  (just going to write the guidelines here). 

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)

# additional tasks

- [x] Implement alpha deploy process to PyPI in GHA
- [x] Added `versionadded` and `versionchanged`
    - [x] still needed for new set like functions for object collections.
- [x] Test unruly strings for material printing
- [x] test set like behavior for numberedobject collection
- [x] promote surfaces to be second or first class objects. 
- [x] test materials with `plib` intermixed with materials
- [ ] ~~document migration.~~ #606
- [x] Create easy way to edit existing Nuclide definition (library), and fraction. 
- [ ] ~~Document all random constant dictionaries in dev stuff, and add to PR checklist~~ #605
- [x] Implement all proposed changes from #475. 
    - [x] implement `mix`ing
- [ ] Manually merge due to immediate deploy
- [x] Cleat out all TODOs
- [x] Document that type strings are now the standard going forward. Mention rolling basis. Mention pytest expectations. 
- [x] Check if setting is_atom_fraction is exported properly is tested. 
- [x] Test cells clone


MicahGale:
Can we remove some old `versionchanged` and `versionadded` that were added in `0.2.0`? I think at least the changes are old enough that they can be treated as the status quoue with 1.0.0

MicahGale:
The benchmark now takes 35 seconds: https://github.com/idaholab/MontePy/actions/runs/12126919015/job/33810033767?pr=507. It looks like the logic is just now more complicated and slower and doesn't seem like there's O(N^2) scaling. I think we should just move the goal post here and deal with optimization later as planned.

MicahGale:
I'm thinking:

1. This is an alpha release. Let's relax
2. Documentation to come as a separate PR into `alpha-test-dev`
3. dictionary documentation is a separate concern from this and needs to be separated out. 

This is all to say that I think I will YOLO merge this once I:

1. implement and test `mix`
2. Clear out/ triage the `TODO` items

MicahGale:
I'm going to YOLO this onto Alpha-test-dev.

PR #528, opened 2024-08-29
Marked material_components for deprecation.

### LINKED ISSUE #506: Deprecation in support of material redesign
Assuming #504  is true and a major release is needed, deprecation warnings are needed. I think `Material.material_components` should raise a `DeprecationWarning` pointing to #504 with release 0.4.1.

Also I think with the next major release a lot of the DeprecationWarnings from Montepy 0.2.0 can be removed now. Just as: `montepy.data_cards`. I think at least the ones that are raised as an Error because changing a `DeprecationWarning` for an `AttributeError` won't be too bad. 

I'm not sure if these new Deprecations should occur before or after a migration plan is in place. The migration plan would need to have the new interface pretty well pinned down, so I'm hesitant to wait. Maybe we could at least make a landing page in the documentation that we can explain the issues, and possible migration, alpha testing etc.

MicahGale:
# Description

See #504.

* Adds deprecation doc strings
* Started a migration page to describe the deprecation and the need for it.
* Raised deprecation warnings with `Material.material_components` is accessed.

Fixes #506 

# Checklist


MicahGale:
Should we switch from the term: `Isotope` to `Nuclide`? This is the terminology used by MCNP, and is meant to be all encompassing or isotopes, metastable isomers, and atomic data. For once I think I agree with their terminology.

tjlaboss:
Yes. Nuclide is better for every reason.

montepy/data_inputs/material.py
```diff
@@ -102,13 +105,23 @@ def is_atom_fraction(self):
...
-        :rtype: dict
+            The keys are :class:`~montepy.data_inputs.isotope.Isotope` instances, and the values are
+            :class:`~montepy.data_inputs.material_component.MaterialComponent` instances.
+
+            :rtype: dict
         """
+        warnings.warn(
+            f"""material_components is deprecated, and will be removed in MontePy 1.0.0.
+See <https://www.montepy.org/migrate0_1.html> for more information """,
+            DeprecationWarning,
```
tjlaboss, line 123: https://stackoverflow.com/q/55377831

I think `DeprecationWarning` is correct

doc/source/users.rst
```diff
@@ -9,5 +9,6 @@ User Guide
    utilities
    tricks
    faq
+   migrate0_1
```
tjlaboss, line 12: Should _migrations/_ be a subfolder?

tjlaboss (commented):
Changes so far look good

doc/source/users.rst
MicahGale: Hopefully there will  be few of these necessary, but  given that these should be permalinks let's do it.

montepy/data_inputs/material.py
MicahGale: I did think about it a little bit. Given this discussion: I don't think there's enough down stream devs that they would see this. If anything I feel like `FutureWarning` might be the safest bet?

tjlaboss (approved):

PR #548, opened 2024-09-15
Remove Python 3.8 support, and add python 3.13 and numpy 2.0 support

### LINKED ISSUE #201: End of Life for python 3.8 support

### LINKED ISSUE #457: Decide how to test against Numpy 1 and 2
Numpy 2.0 came out this June and is the first major release in a long time, which means some backwards compatibility is lost. Numpy 2.0 requires python 3.9. Due to this I saw that for python 3.8 we are testing against Numpy 1.24, and for Python >=3.9 we are testing against numpy 2.0. So right now we are accidentally testing against both major versions. However, this isn't a good spot to be.

So actions IMO:
1. Research what breaks in 2.0
2. Determine if this matters to us.
3. Decide if we need to peg to Numpy 1.xxx or Numpy 2.xxx
4. Decide if we need to do matrix testing against both major versions of numpy.

Thoughts @tjlaboss?

MicahGale:
# Description

Fixes #201, fixes #457

# Checklist

- [x] Remove fail-fast flag.


MicahGale:
Want to review this now @tjlaboss, so if everything passes tomorrow we can merge on Python day?

MicahGale:
It looks like pytest-profile is incompatible with 3.13. I'm thinking of just uninstalling it for now as a work around

tjlaboss:
Yeah, we can remove it from CI and just profile <=3.12  locally

MicahGale:
We can still profile. Just do the profiling on <= 3.12 in CI. We don't need profiling to run the test suite in 3.13

tjlaboss (commented):
LGTM pending Python 3.13 CI

tjlaboss (approved):

PR #554, opened 2024-09-24
Made multi-format multi-resolution favicons.

MicahGale:
# Description

I noticed that on DuckDuckGo our favicon doesn't show up in the results page:

image

Looking into how Bing handles this (most of the backend for DuckDuckGo), this is because we only provide an SVG icon. [This forum post](https://answers.microsoft.com/en-us/bing/forum/all/what-are-the-requirements-to-display-favicon-on/236d6598-41e2-483c-9950-aae8714d943c) recommends providing a PNG, and including it in the sitemap. I couldn't get it to work in the sitemap though, so I'm ignoring that, and [SO doesn't say this is necessary](https://stackoverflow.com/questions/64526263/how-to-show-my-website-favicon-in-bings-search-engine). 

[Another SO thread recommends providing lots of options](https://stackoverflow.com/questions/48956465/favicon-standard-2024-svg-ico-png-and-dimensions). I decided we should do most of those. I couldn't get inkscape to export ICO so I think we can just ignore IE support. The new extension does work in adding multiple favicon options:

image
 
It is intentionally ordered from best to worst options. 

# Checklist


tjlaboss (changes requested):
I added a 32x32 `ico`.  Add that to the docs and then we'll merge

MicahGale:
> I added a 32x32 `ico`. Add that to the docs and then we'll merge

Added!

tjlaboss (approved):

PR #539, opened 2024-09-10
Implementing weakref

MicahGale:
# Description

Please include a summary of the change and which issue is fixed if applicable. Please also include relevant motivation and context.

Fixes #514 

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)

PR #563, opened 2024-10-02
remove o(n^2) operation from NumberedObjectCollection

### LINKED ISSUE #556: Remove O(N^2) operation from NumberedObjectCollection

I was analyzing some data from parsing times and found that we have O(N^2) scaling despite fixing #510. After some investigation I found the culprit is in `NumberedObjectCollections.append`. This runs through all of `self.numbers` every time which is a generator over the entire list.... O(N^2).

So this should be eliminated.

A few solutions.

1. Remove this operation from `read_input` and do a bulk add that is optimized
2. Make `NumberedObjectCollection` a lot more like dicts, and rely on `__num_cache` a lot more. Through the number property we guarantee that numbers are always update for objects tied to problem. So maybe we use this O(n^2) operation only as a last resot.

MicahGale:
# Description

As described in #556 there was an O(N^2) operation for appending N items to a `NumberedObjectCollection`. This was because `self.numbers` (which is an O(N) generator) was checked every append. This was fixed by:

1. Moving all number conflict checks to the existing `check_number`
2. Trusting `__num_cache` explicitly iff this problem is linked to a problem.
3. creating a method: `_update_number` that updates `__num_cache` appropriately when a number is changed. 
4. Ensuring all `Numbered_MCNPObject` instances run `_update_number` if they are attached a problem. This is why `__num_cache` is trusted in this case.

This doesn't remove all bottlenecks and get us to O(N) time, but it does make `append` no longer a bottle neck as can be seen in the profiling results. 

Don't review this until #539 is merged because this is built off of that branch. 

Fixes #556

# Checklist


MicahGale:
This depends on #539

MicahGale:
There is still an O(N^2) call to `.numbers` for some reason.

MicahGale:
I can't replicate this O(N^2) behavior. I wonder if the wrong version of MontePy was used in the test.

MicahGale:
Upon further examination this was a problem with my tests. If you look at the profile results in the CI, `numbers` doesn't even show up and `append` is very low on the list. So I think we can say this was solved.

montepy/numbered_object_collection.py
```diff
@@ -128,11 +127,27 @@ def check_number(self, number):
...
+        else:
+            if number in self.numbers:
+                conflict = True
+        if conflict:
             raise NumberConflictError(
                 f"Number {number} is already in use for the collection: {type(self)} by {self[number]}"
             )
 
+    def _update_number(self, old_num, new_num, obj):
+        """ """
```
tjlaboss, line 144: empty docstring

montepy/numbered_object_collection.py
```diff
@@ -177,22 +192,25 @@ def extend(self, other_list):
         """
         if not isinstance(other_list, (list, type(self))):
             raise TypeError("The extending list must be a list")
+        if self._problem:
+            nums = set(self.__num_cache.keys())
```
tjlaboss, line 196:
```python
nums = set(self.__num_cache)
```

tjlaboss (changes requested):

tjlaboss:
Two small changes to make.

montepy/numbered_object_collection.py
MicahGale: I like those.

MicahGale:
No change in memory usage (duh), but went from 27.6 seconds to 26.7 seconds for this improvement. This is a 3% reduction for the "big model". Though this is a low cost O(N^2) operation, and the saving should scale accordingly.

tjlaboss (approved):

Issue #558, opened 2024-09-30
Bug when attempting to edit a cell geometry


MicahGale:
**To Reproduce**

A short code snippet of what you have ran. Please change or remove any specific values or anything that can't be public. For example:

``` python
input = montepy.input_parser.mcnp_input.Input(["1 0 (-1 : 2) 3 imp:n=1"], montepy.input_parser.block_type.BlockType.CELL)
cell = montepy.Cell(input)
surf4 = montepy.surfaces.surface.Surface()
surf4.number = 4
surf5 = montepy.surfaces.surface.Surface()
surf5.number = 5
cell.geometry = (cell.geometry & -surf4) | +surf5
```

**Error Message (if any)**

If an error message was printed please include the entire stacktrace. If it includes any specific values please change or remove them. For example:

``` python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 cell.geometry = (cell.geometry & -surf4) | +surf5

File ~/mambaforge/lib/python3.11/site-packages/montepy/utilities.py:160, in make_prop_pointer.<locals>.decorator.<locals>.setter(self, value)
    158     value = base_type(value)
    159 if validator:
--> 160     validator(self, value)
    161 setattr(self, hidden_param, value)

File ~/mambaforge/lib/python3.11/site-packages/montepy/cell.py:30, in _link_geometry_to_cell(self, geom)
     28 def _link_geometry_to_cell(self, geom):
     29     geom._cell = self
---> 30     geom._add_new_children_to_cell(geom)

File ~/mambaforge/lib/python3.11/site-packages/montepy/surfaces/half_space.py:195, in HalfSpace._add_new_children_to_cell(self, other)
    193 for item in container:
    194     if item not in parent:
--> 195         parent.append(item)

File ~/mambaforge/lib/python3.11/site-packages/montepy/numbered_object_collection.py:248, in NumberedObjectCollection.append(self, obj)
    241 """Appends the given object to the end of this collection.
    242
    243 :param obj: the object to add.
    244 :type obj: Numbered_MCNP_Object
    245 :raises NumberConflictError: if this object has a number that is already in use.
    246 """
    247 if not isinstance(obj, self._obj_class):
--> 248     raise TypeError(f"object being appended must be of type: {self._obj_class}")
    249 if obj.number in self.numbers:
    250     raise NumberConflictError(
    251         (
    252             "There was a numbering conflict when attempting to add "
[...]

Issue #552, opened 2024-09-19
ListNode comparison is blocking contains for ListNode


MicahGale:
**Describe the bug**

**To Reproduce**

A short code snippet of what you have ran. Please change or remove any specific values or anything that can't be public. For example:

``` python
problem = montepy.read_input("foo.imcnp")
```

**Error Message (if any)**

If an error message was printed please include the entire stacktrace. If it includes any specific values please change or remove them. For example:

``` python
MontePy/montepy/input_parser/input_reader.py:31: in read_input
    problem.parse_input(replace=replace)
MontePy/montepy/mcnp_problem.py:282: in parse_input
    obj = obj_parser(input)
MontePy/montepy/cell.py:100: in __init__
    self._parse_keyword_modifiers()
MontePy/montepy/cell.py:124: in _parse_keyword_modifiers
    input = input_class(in_cell_block=True, key=key, value=value)
MontePy/montepy/data_inputs/fill.py:46: in __init__
    self._parse_cell_input(key, value)
MontePy/montepy/data_inputs/fill.py:115: in _parse_cell_input
    if "(" in data.nodes:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = (shortcut:Shortcuts.REPEAT: deque([(Value, 2.0, padding: (Padding, [' '])), (Value, 2.0, padding: (Padding, [' '])), (...e, 2.0, padding: (Padding, [' '])), (Value, 2.0, padding: (Padding, [' '])), (Value, 2.0, padding: (Padding, [' ']))]))
other = '('

    def __eq__(self, other):
        if not isinstance(other, (type(self), list)):
>           raise TypeError(
                f"ListNode can only be compared to a ListNode or List. {other} given."
            )
E           TypeError: ListNode can only be compared to a ListNode or List. ( given.

MontePy/montepy/input_parser/syntax_node.py:1703: TypeError
```

**MCNP input file snippet**

See: `tests/models/icsbep/HEU-COMP-THERM-021_case_007.imcnp`

**Version**

 - Version 0.4.1

Issue #551, opened 2024-09-19
E0 not allowed by Parser


MicahGale:
**Describe the bug**

While playing around with a huge set of MCNP input files I found this error. I suspect it is due to the fact that the `ParserBase` distinguishes between 0 and non-0 numbers. I think this comes from [this line](https://github.com/idaholab/MontePy/blob/develop/montepy/input_parser/parser_base.py#L460).

**To Reproduce**

A short code snippet of what you have ran. Please change or remove any specific values or anything that can't be public. For example:

``` python
problem = montepy.read_input("foo.imcnp")
```

**Error Message (if any)**

If an error message was printed please include the entire stacktrace. If it includes any specific values please change or remove them. For example:

``` python
MontePy/montepy/input_parser/input_reader.py:31: in read_input
    problem.parse_input(replace=replace)
MontePy/montepy/mcnp_problem.py:297: in parse_input
    raise e
MontePy/montepy/mcnp_problem.py:282: in parse_input
    obj = obj_parser(input)
MontePy/montepy/data_inputs/data_parser.py:42: in parse_data
    base_input = data_input.DataInput(input, fast_parse=True)
MontePy/montepy/data_inputs/data_input.py:340: in __init__
    super().__init__(input, fast_parse)
MontePy/montepy/data_inputs/data_input.py:64: in __init__
    super().__init__(input, self._classifier_parser)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <montepy.data_inputs.data_input.DataInput object at 0x7f27acb0b080>
input = INPUT: BlockType.DATA: ['E0  1.0000E-10 1.0500E-10 1.1000E-10 1.1500E-10 1.2000E-10', '      1.2750E-10 1.3500E-10 1.4...                                   ***', 'c *************************************************************************']
parser = <montepy.input_parser.data_parser.ClassifierParser object at 0x7f27f003bce0>

    def __init__(self, input, parser):
        self._problem = None
        self._parameters = ParametersNode()
        self._input = None
        if input:
            if not isinstance(input, montepy.input_parser.mcnp_input.Input):
[...]

PR #567, opened 2024-10-09
Release 0.5.0

MicahGale:
# Features Added

* Added `clone` method to simplify making copies of objects (#469).

# Performance Improvement

* Fixed cyclic memory reference that lead to memory leak in `copy.deepcopy` (#514).
* Fixed O(N<sup>2</sup>) operation in how append works for object collections like Cells (#556).

# Bug Fixes

* Fixed bug with parsing an `EO` input (#551).
* Fixed a bug raised in an edge case when editing cell geometry, by making the error clearer (#558).
* Fixed bug with having a shortcut in a cell fill (#552).
* Fixed bug where file streams couldn't actually be read (#553).

# Support

* Added support for Python 3.13, and removed support for Python 3.8, and officially added support NumPy 1 & 2 (#548).


tjlaboss:
Need to update the changelog and then we can release.

MicahGale:
The changelog here?

MicahGale:
Ohh I see the issue you are referring to I think. A section header got deleted.

MicahGale:
NVM. The `#Next Version#` thing. That gets done manually by me with `.github/scripts/deploy.sh`. It's a bit jank but it works for now.

PR #568, opened 2024-10-15
Added hierarchy to MontePy changelog.

MicahGale:
# Description

Quick change to how the changelog is organized to avoid having dozens of top level entries. Added organization at the minor release level. This looks like:

image

This will be helpful one day if we need to split the changelog into multiple files. 
# Checklist


tjlaboss (approved):
Nice.

PR #641, opened 2025-01-22
Added autodocsumm to make TOCs in docs

MicahGale:
This is a quick improvement that makes it so that a class and modules have a TOC at the top the documentation to quickly find methods and attributes.


MicahGale:
Here is a good demo: https://montepy--641.org.readthedocs.build/en/641/api/montepy.cell.html

tjlaboss (approved):
is good

PR #663, opened 2025-02-13
Allow strictly finding material components

### LINKED ISSUE #642: Search for Material Nuclides containing just elemental data.

Sometimes you get a model with legacy nuclear data, e.g., `26000.50c` that you want to replace.
However, if you do `problem.materials.get_containing("26000.50c")` you will get all iron isotopes.
It's unclear right now if the library extension is being matched (needs verification).

While in some cases this is a nice behavior, its not always the best. 

I think an `strict` parameter would be helpful.

So `problem.materials.get_containing("26000.50c", strict=True)` would exactly match the required nuclide.
This would also be helpful for finding library-less nuclides. 

The alternatives might be:

``` python

for material in problem.materials:
    for nuclide in material.nuclides:
        if nuclide == "2600.50c": 
             yield material
             break
```

**Additional context**
This is for version 1.0.0a1.

MicahGale:

### Description

This is a feature implementation for `contains`, `get_containing`, and `find`. The topline is that there is a new `strict` argument that switches from implicit to explicit matching. So with `strict=True` `"H"` only matches `1000` and not `1001.80c` nor `1000.80c`. The other main thing was adding some feedback from @gonuke.

This feedback splint `contains` (and `get_containing`) to be: `contains_all` and `contains_any`. This should be clearer to users, and be more useful. The main use case of `contains_any` might be:

``` python
fissile = list(problem.material.get_containing_any("U-235", "U-233", "Pu-239",...)
```

Fixes #642

---

### General Checklist

---

<details open> 

- [ ] ~For infrastructure updates, I have updated the developer's guide.~
- [x] For significant new features, I have added a section to the getting started guide.

</details>

---

<details>

</details>

---

Ensure that:

----
📚 Documentation preview 📚: https://montepy--663.org.readthedocs.build/en/663/

### DIFF
--- a/montepy/data_inputs/material.py
+++ b/montepy/data_inputs/material.py
@@ -724,6 +724,12 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
 
         .. note::
 
+            The difference between :func:`contains_all` and :func:`contains_any` is only for how they
+            handle being given multiple nuclides. This does not impact how given Elements will match 
+            daughter Nuclides. This is handled instead by ``strict``.
+        
+        .. note::
+
             For details on how to use the ``strict`` argument see the examples in: :func:`find`.
 
         .. versionadded:: 1.0.0
--- a/montepy/data_inputs/material.py
+++ b/montepy/data_inputs/material.py
@@ -726,11 +726,6 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
 
             For details on how to use the ``strict`` argument see the examples in: :func:`find`.
 
-        .. note::
-
-            If a nuclide is in a material multiple times, and cumulatively exceeds the threshold,
-            but for each instance it appears it is below the threshold this method will return False.
-
         .. versionadded:: 1.0.0
 
         :param nuclides: a plurality of nuclides to check for.
@@ -781,11 +776,6 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
         .. note::
 
             For details on how to use the ``strict`` argument see the examples in: :func:`find`.
-
-        .. note::
-
-            If a nuclide (or element) is in a material multiple times, and cumulatively exceeds the threshold,
-            but for each instance it appears it is below the threshold this method will return False.
 
         .. versionadded:: 1.0.0
 
@@ -857,27 +847,26 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
         element_search = {}
         for nuclide in nuclide_finders:
             if isinstance(nuclide, Element):
-                element_search[nuclide] = False
+                element_search[nuclide] = 0.0
             if isinstance(nuclide, Nucleus):
-                nuclei_search[nuclide] = False
+                nuclei_search[nuclide] = 0.0
             if isinstance(nuclide, Nuclide):
-                nuclides_search[str(nuclide).lower()] = False
+                nuclides_search[str(nuclide).lower()] = 0.0
 
         for nuclide, fraction in self:
-            if fraction < threshold:
-                continue
             if str(nuclide).lower() in nuclides_search:
-                nuclides_search[str(nuclide).lower()] = True
+                nuclides_search[str(nuclide).lower()] += fraction
             if nuclide.nucleus in nuclei_search:
-                nuclei_search[nuclide.nucleus] = True
+                nuclei_search[nuclide.nucleus] += fraction
             if nuclide.element in element_search:
-                element_search[nuclide.element] = True
+                element_search[nuclide.element] += fraction
 
+        threshold_check = lambda x: x > threshold
         return bool_func(
             (
-                bool_func(nuclides_search.values()),
-                bool_func(nuclei_search.values()),
-                bool_func(element_search.values()),
+                bool_func(map(threshold_check, nuclides_search.values())),
+                bool_func(map(threshold_check, nuclei_search.values())),
+                bool_func(map(threshold_check, element_search.values())),
             )
         )
 
--- a/montepy/data_inputs/material.py
+++ b/montepy/data_inputs/material.py
@@ -765,7 +765,7 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
         Checks if this material contains any of the given nuclide.
 
         A boolean "or" is used for this comparison.
-        That is this material must contain all nuclides at or above the given threshold
+        That is, this material must contain any nuclides at or above the given threshold
         in order to return true.
 
         Examples
@@ -786,7 +786,7 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
 
         .. note::
 
-            If a nuclide is in a material multiple times, and cumulatively exceeds the threshold,
+            If a nuclide (or element) is in a material multiple times, and cumulatively exceeds the threshold,
             but for each instance it appears it is below the threshold this method will return False.
 
         .. versionadded:: 1.0.0
--- a/montepy/data_inputs/material.py
+++ b/montepy/data_inputs/material.py
@@ -827,7 +827,7 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
     def _contains_arb(
         self,
         *nuclides: Union[Nuclide, Nucleus, Element, str, int],
-        bool_func: co.abc.Callable[co.abc.Iterable[bool]] = all,
+        bool_func: co.abc.Callable[co.abc.Iterable[bool]] = None,
         threshold: float = 0.0,
         strict: bool = False,
     ) -> bool:
--- a/montepy/data_inputs/material.py
+++ b/montepy/data_inputs/material.py
@@ -807,7 +807,34 @@ See <https://www.montepy.org/migrations/migrate0_1.html> for more information ""
         :raises ValueError: if the fraction is not positive or zero, or if nuclide cannot be interpreted as a Nuclide.
 
         """
-        nuclides = []
+        return self._contains_arb(
+            *nuclides, bool_func=any, threshold=threshold, strict=strict
+        )
+
+    @staticmethod
+    def _promote_nuclide(nuclide, strict):
+        if not isinstance(nuclide, (str, int, Element, Nucleus, Nuclide)):
+            raise TypeError(
+                f"Nuclide must be a type that can be converted to a Nuclide. The allowed types are: "
+                f"Nuclide, Nucleus, str, int. {nuclide} given."
+            )
+        if isinstance(nuclide, (str, int)):
+            nuclide = Nuclide(nuclide)
+        # treat elemental as element
+        if isinstance(nuclide, (Nucleus, Nuclide)) and nuclide.A == 0 and not strict:
+            nuclide = nuclide.element
+        if isinstance(nuclide, Nuclide) and not str(nuclide.library) and not strict:
+            nuclide = nuclide.nucleus
+        return nuclide
+
+    def _contains_arb(
+        self,
+        *nuclides: Union[Nuclide, Nucleus, Element, str, int],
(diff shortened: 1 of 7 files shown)


MicahGale:
TODO: `find`

montepy/data_inputs/material.py
```diff
@@ -742,6 +748,9 @@ def contains(
         :param threshold: the minimum concentration of a nuclide to be considered. The material components are not
             first normalized.
         :type threshold: float
+        :param strict: Whether to not let an elemental nuclide match all child isotopes, isomers, not have an isotope
```
MicahGale, line 751: Make this explicit of what is True/False.

montepy/data_inputs/material.py
```diff
@@ -729,10 +730,15 @@ def contains(
...
+            For details on how to use the ``strict`` argument see the examples in: :func:`find`.
+
         .. note::
 
             If a nuclide is in a material multiple times, and cumulatively exceeds the threshold,
             but for each instance it appears it is below the threshold this method will return False.
+
         .. versionadded:: 1.0.0
 
         :param nuclide: the first nuclide to check for.
```
MicahGale, line 744: Combine to be `*nucides`

MicahGale: Fissile example is bad. It is `&` not `|`

MicahGale: Make `contains_all` and `contains_any`

montepy/data_inputs/material.py
```diff
@@ -760,9 +769,13 @@ def contains(
             if isinstance(nuclide, (str, int)):
                 nuclide = Nuclide(nuclide)
             # treat elemental as element
-            if isinstance(nuclide, (Nucleus, Nuclide)) and nuclide.A == 0:
+            if (
```
MicahGale, line 772: Out of order. Do type checking for all first.

montepy/data_inputs/material.py
```diff
@@ -760,9 +769,13 @@ def contains(
             if isinstance(nuclide, (str, int)):
```
MicahGale, line 769: Make a `is_nuclide` or `nuclide_promoter` internal function for all these calls.

montepy/data_inputs/material.py
```diff
@@ -750,38 +807,57 @@ def contains(
...
+        if isinstance(nuclide, (Nucleus, Nuclide)) and nuclide.A == 0 and not strict:
+            nuclide = nuclide.element
+        if isinstance(nuclide, Nuclide) and not str(nuclide.library) and not strict:
+            nuclide = nuclide.nucleus
+        return nuclide
 
+    def _contains_arb(
+        self,
+        *nuclides: Union[Nuclide, Nucleus, Element, str, int],
+        bool_func: co.abc.Callable[co.abc.Iterable[bool]] = all,
```
tjlaboss, line 833: I don't think `bool_func` should have a default.

montepy/data_inputs/material.py
```diff
@@ -750,38 +807,57 @@ def contains(
...
-            if nuclide not in self:
-                return False
+        if bool_func == all:
+            for nuclide in nuclide_finders:
+                if nuclide not in self and bool_func == all:
+                    return False
 
         nuclides_search = {}
         nuclei_search = {}
         element_search = {}
```
tjlaboss, line 859: Since we only ever use `values()`, these don't need to be dicts. We can actually get away with just a set.

The dicts make it easier to maintain and debug though, and the performance penalty is very minor.

montepy/materials.py
```diff
@@ -30,26 +31,75 @@ class Materials(NumberedDataObjectCollection):
...
+        :raises TypeError: if any argument is of the wrong type.
+        :raises ValueError: if the fraction is not positive or zero, or if nuclide cannot be interpreted as a Nuclide.
+        """
+        return self._contains_arb(
+            *nuclides, bool_func=any, threshold=threshold, strict=strict
+        )
+
+    def get_containing_all(
+        self,
+        *nuclides: Union[
```
tjlaboss, line 91: *args are necessarily a tuple.

montepy/data_inputs/material.py
```diff
@@ -692,16 +692,16 @@ def add_nuclide(self, nuclide: Union[Nuclide, str, int], fraction: float):
             nuclide = Nuclide(nuclide)
         self.append((nuclide, fraction))
 
-    def contains(
+    def contains_all(
         self,
-        nuclide: Union[Nuclide, Nucleus, Element, str, int],
-        *args: Union[Nuclide, Nucleus, Element, str, int],
+        *nuclides: Union[Nuclide, Nucleus, Element, str, int],
```
tjlaboss, line 697: *args are necessarily a tuple:

```python
Tuple[Union[Nuclide, Nucleus, Element, str, int], ...]
```

This type hint is long and frequently reused. Store it as a variable and reuse it.

montepy/materials.py
```diff
@@ -30,26 +31,75 @@ class Materials(NumberedDataObjectCollection):
     def __init__(self, objects=None, problem=None):
         super().__init__(Material, objects, problem)
 
-    def get_containing(
+    def get_containing_any(
         self,
-        nuclide: Union[
+        *nuclides: Union[
```
tjlaboss, line 36: *args are necessarily a tuple

montepy/materials.py
```diff
@@ -72,39 +122,41 @@ def get_containing(
...
-            if isinstance(nuclide, (str, int)):
-                nuclide = montepy.Nuclide(nuclide)
-            nuclides.append(nuclide)
+        return self._contains_arb(
+            *nuclides, bool_func=all, threshold=threshold, strict=strict
+        )
+
+    def _contains_arb(
+        self,
+        *nuclides: Union[
```
tjlaboss, line 146: *args are necessarily a tuple

tjlaboss (changes requested):
Without running the code: the logic makes sense and the tests make sense. 

I will give this function a test drive and confirm that everything works as expected before approving. In the meanwhile,

The type hinting for `*nuclides* is incorrect and one of the docstrings needs to be fixed. There are a few optional things to consider.

montepy/data_inputs/material.py
MicahGale: Should I re-order the args or would a `None` default be good? In the current position a default is mandatory.

montepy/data_inputs/material.py
MicahGale: HOW DID I MISS THE CHANCE FOR SETS? I'll double check if this actually viable.

montepy/materials.py
MicahGale: But this begs the question: with `*args` is it clearer to document one instance or the set?  I argue that `Tuple` will confuse users and make them try: `get_containing_all(("U-238", "U-235"))`

montepy/data_inputs/material.py
tjlaboss: Oh, right: because it proceeds `*nuclides`. It's not important.

montepy/data_inputs/material.py
tjlaboss: perhaps somthing like `T_Nuclidelike`

montepy/materials.py
tjlaboss: You're right, PEP-484 says to document the instance (not the tuple): https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

montepy/data_inputs/material.py
```diff
@@ -715,33 +715,90 @@ def contains(
...
 
-            # try to find a uranium
+        .. note::
+
+            For details on how to use the ``strict`` argument see the examples in: :func:`find`.
+
+        .. note::
+
+            If a nuclide is in a material multiple times, and cumulatively exceeds the threshold,
+            but for each instance it appears it is below the threshold this method will return False.
```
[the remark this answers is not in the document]
MicahGale: Reason: laziness of implementation. Do you think it's worth doing the sum? For elements I could see this being worthwhile.

montepy/data_inputs/material.py
MicahGale: Ok I don't think sets are viable; especially if we want to track this cumulatively.

montepy/data_inputs/material.py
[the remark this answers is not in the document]
tjlaboss: Yes, it is worth doing the sum: principle of least astonishment. 

The documentation should also disambiguate:

```python
mat.contains_all("U", strict=False)
```

It is intuitive that `any` and `all` are interchangeable an argument of 1 item, but there is the potential for confusion when `strict=False` because then the `Element` represents `any` isotope thereof.

montepy/data_inputs/material.py
tjlaboss: Furthermore, `bool_func` is arbitrary, so it doesn't necessarily _have_ to be just `any` or `all`. I'll mark as resolved.

montepy/data_inputs/material.py
```diff
@@ -750,38 +807,57 @@ def contains(
...
+        *nuclides: Union[Nuclide, Nucleus, Element, str, int],
+        bool_func: co.abc.Callable[co.abc.Iterable[bool]] = all,
+        threshold: float = 0.0,
+        strict: bool = False,
+    ) -> bool:
+        nuclide_finders = []
         if not isinstance(threshold, float):
             raise TypeError(
                 f"Threshold must be a float. {threshold} of type: {type(threshold)} given"
             )
```
tjlaboss, line 838: Should've caught this before (my bad):

Do not enforce that a numeric value is a `float` type. This will disallow everything from `np.float16` to `int`. Instead, assert that it is an instance of `numbers.Number`, or better yet, do not enforce: it will fail naturally on the next line if it cannot be compared to 0.

montepy/materials.py
MicahGale: Ok this sequences won't reliable show a nice doc string. So I'm thinking of doing PEP 484.

montepy/data_inputs/material.py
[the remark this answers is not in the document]
MicahGale: Ok I'll sum. I'm not sure exactly what you want with that disambiguation though.

montepy/data_inputs/material.py
MicahGale: We need to make an issue to do this across the whole code base.

MicahGale: Also not going to let this fall through. Users will get confused and think MontePy is the problem and not them.

montepy/data_inputs/material.py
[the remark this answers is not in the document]
tjlaboss: State that `contains_all` with an `Element` and `strict=False` will match any `Nuclide` of that element (isotope).

just to be abundantly clear that by `all` it doesn't mean all possible isotopes.

montepy/data_inputs/material.py
tjlaboss: I will open an issue

tjlaboss:
Test now failing on: [material.py#L555-L556](https://github.com/idaholab/MontePy/blob/67e38bbd865c0e860bfbaa67368a7defe0bced3e/montepy/data_inputs/material.py#L555-L556)

```python
        if not isinstance(newvalue[0], Nuclide):
            raise TypeError(f"First element must be an Nuclide. {newvalue[0]} given.")
```

tjlaboss:
It is given a `Nucleus`, not a `Nuclide`. Should this be changed to the new `NuclideLike`?

MicahGale:
> It is given a `Nucleus`, not a `Nuclide`. Should this be changed to the new `NuclideLike`?

I broke something with switching to `_promote_nuclide` in `add_nuclide`. This function can return a `Nucleus` or `Element` sometimes if `strict` is `False`.

tjlaboss (approved):
LGTM.

Issue #745, opened 2025-04-14
Issues with exporting models on Windows


MicahGale:
## Bug Description

This bug came up while testing #743: The test suite when ran on windows fails. Specifically it appears to be the tests that test exporting an MCNP model. Some of the failures are benign, like shortcuts not recompressing, but some of the test failures suggest a lost of information. 

## To Reproduce

 

On windows run: 
``` bash
pytest
```

## Error Message 

<details > 

``` python
============================= test session starts =============================
platform win32 -- Python 3.13.2, pytest-8.3.5, pluggy-1.5.0
rootdir: D:\a\MontePy\MontePy
configfile: pyproject.toml
plugins: hypothesis-6.131.0, profiling-1.8.1
collected 95 items

tests\test_integration.py ............F................................. [ 48%]
......................................F......FF..                        [100%]

================================== FAILURES ===================================
______________________________ test_problem_str _______________________________

simple_problem = MCNP problem for: tests\inputs\test.imcnp
MESSAGE: 3 lines
TITLE: MCNP Test Model for MOAA
CELL: 1, mat: 1, DENS: 20.0...src 
DATA INPUT: kcode 
DATA INPUT: phys:p 
Mode: {<Particle.NEUTRON: 'N'>, <Particle.PHOTON: 'P'>}
vol NO 2J 1 1.5 J



    def test_problem_str(simple_problem):
        output = str(simple_problem)
>       assert "MCNP problem for: tests/inputs/test.imcnp" in output
E       AssertionError: assert 'MCNP problem for: tests/inputs/test.imcnp' in 'MCNP problem for: tests\\inputs\\test.imcnp, TITLE: MCNP Test Model for MOAA'

tests\test_integration.py:190: AssertionError
________________________ test_read_write_cycle[file6] _________________________

file = WindowsPath('tests/inputs/test_interp_edge.imcnp')

    @pytest.mark.parametrize(
        "file",
        set((Path("tests") / "inputs").iterdir())
        - {
            Path("tests")
            / "inputs"
            / p  #                           Skip complexity of read
            for p in constants.BAD_INPUTS
            | constants.IGNORE_FILES
            | {"testRead.imcnp", "readEdgeCase.imcnp"}
        },
    )
    def test_read_write_cycle(file):
        print(f"Testing against {file} *********************")
        if ".swp" in file.suffixes:
            return
        problem = montepy.read_input(file, multi_proc=MULTI_PROC)
[...]

PR #749, opened 2025-04-25
Feature/361 find cells generator

### LINKED ISSUE #361: Add generator to find cells filled with a universe

I was trying to find if a universe was included in a problem and found it wasn't easy to 

I would like a generator ala `universe.cells`. Maybe something `universe.fill_cells`

What I did instead:

``` python
for cell in problem.cells:
   if cell.fill:
      if cell.fill.universe is universe:
         ...
```

kordusas:

### Description

Implemented a generator to return the cells which are the universe filled with. It flattens the lattice and looks through if any of universes in the lattice match. 

- Created an MCNP input with lattice and fill to test
- developed test case for the input using pytest 
- Migrated universe tests from unittest to pytest
- ~~created dundermethods~~ `__eq__` ~~for universe objects~~

Fixes #361. 

---

### General Checklist

---

<details open> 

- [x] ~~For infrastructure updates, I have updated the developer's guide.~~
- [x] ~~For significant new features, I have added a section to the getting started guide.~~

</details>

---

Ensure that:


kordusas:
If you agree with the suggested changes I could also update the testing to follow pytest instead of unitest for the universe.

MicahGale:
Thanks for doing this. I probably won't be able to review it this week due to M&C though.

MicahGale:
> If you agree with the suggested changes I could also update the testing to follow pytest instead of unitest for the universe.
> 
> I'm seeing some tests fail even on the unmodified develop branch when running on Windows. It looks like line ending differences (`\r\n` vs `\n`) may be causing it. should i try to normalize this so tests pass cross-platform?

Yes please, if you can! This was frustrating me with #745.

MicahGale:
I'm digging into the test failures:

1. "add RTD links" this seems to be a permission issue with non-maintainer PRs. This is on me.
2. "doc-build" this is `osti.gov` links failing. In the past we have seen that the Github CI machines seem to be banned from osti.gov, and this is becoming the status quo (we probably caused this). Could you add osti to the list of ignored URLs? https://github.com/idaholab/MontePy/blob/develop/doc/source/conf.py#L77
3. "doc-test" it looks like you have changed something that is breaking the example code in the documentation. You can test this locally with `make doctest` in the `docs` folder.

montepy/universe.py
```diff
@@ -47,6 +49,24 @@ def cells(self):
                 if cell.universe == self:
                     yield cell
 
+    @property
+    def filled_cells(self):
```
MicahGale, line 53: I think I like this property after a bit of thought. Do you like it @tjlaboss?

We are moving to adding type annotations to everything:

``` suggestion
    @property
    def filled_cells(self) -> Generator[cell]:
```

Have I added this to the developer guide?

montepy/universe.py
```diff
@@ -47,6 +49,24 @@ def cells(self):
...
+        """A generator of the cells that use this universe.
+
+        Returns
+        -------
+        Generator[Cell]
+            an iterator of the Cell objects which use this.
```
MicahGale, line 59:
``` suggestion
            an iterator of the Cell objects which use this universe as their fill.
```

I think it might be good to add a note that clarifies that it considers both `fill.universe` and `fill.universes`.

montepy/universe.py
```diff
@@ -47,6 +49,24 @@ def cells(self):
...
+        """
+        if self._problem:
+            for cell in self._problem.cells:
+                if cell.fill:
+                    if cell.fill.universes is not None:
+                        if np.any(cell.fill.universes.flatten() == self):
+                            yield cell
+                    elif cell.fill.universe == self:
+                        yield cell
+
```
MicahGale, line 69: Should make sure this behaves like a generator even with there is nothing to yield.

I think you can do something like:

``` suggestion
                        yield cell
        else:
            yield from []
```

Though @tjlaboss  would prefer if this were flatter and instead do something like:

``` python
@property 
def filled_cells():
     if not self._problem:
          yield from []
          return
    for cell in self._problem.cells:
         ...
```

tests/inputs/test_lattice_fill_1.imcnp
```diff
@@ -0,0 +1,70 @@
+Made up PWR pin Cell Problem
+c ------------------------
+c Fuel Rod U=1
```
MicahGale, line 3: Is there a reason something like [`tests/inputs/test_universe.imcnp`](https://github.com/idaholab/MontePy/blob/develop/tests/inputs/test_universe.imcnp) doesn't suffice?

tests/test_universe.py
```diff
@@ -1,9 +1,10 @@
 # Copyright 2024 - 2025, Battelle Energy Alliance, LLC All Rights Reserved.
 from hypothesis import given, strategies as st
 import pytest
-from unittest import TestCase
-
+import numpy as np
+import os
 import copy
```
MicahGale, line 6: These blocks could be combined and alphabetized. But this is a mess I made so not a priority.

doc/source/changelog.rst
```diff
@@ -5,6 +5,15 @@ MontePy Changelog
 1.0 releases
 ============
 
+1.0.1
```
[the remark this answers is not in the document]
MicahGale: Also this would be release `1.1.0` since this is adds a feature, and isn't just a bug fix.

tests/test_universe.py
```diff
@@ -38,17 +38,17 @@ def setUp(self):
...
-        self.assertTrue(not card.not_truncated)
+        assert card.old_number == 5
+        assert not card.not_truncated
         # test bad float
-        with self.assertRaises(ValueError):
+        with pytest.raises(ValueError):
             tree = copy.deepcopy(self.tree)
             tree["data"].nodes.pop()
             tree["data"].append(syntax_node.ValueNode("5.5", float))
             card = UniverseInput(in_cell_block=True, key="U", value=tree)
```
MicahGale, line 48: How is `self.tree` allocated if `setup` is no longer being called.

tests/test_universe.py
```diff
@@ -245,166 +245,153 @@ def setUp(self):
...
+            ("1 0 -1 fill=0:1 hi:1 0:1 hi", ValueError),  # "String index"
+            ("1 0 -1 fill=0:1 0:1 0:1 -1", ValueError),  # "Negative universe"
+            (
+                "1 0 -1 fill=0:1 1:0 0:1 1 2 3 4 5 6 7 8",
+                ValueError,
+            ),  # "Inverted bounds"
+            ("1 0 -1 fill=0:1 0:1.5 0:1 1 2 3 4 5 6 7 8", ValueError),  # "Float bounds"
+        ],
+    )
+    def test_complicated_fill_init_error(self, input_str, expected_error):
```
MicahGale, line 321: Nicely done.

tests/test_universe.py
```diff
@@ -245,166 +245,153 @@ def setUp(self):
...
-        list_node.append(syntax_node.ValueNode("5", float))
-        value = syntax_node.SyntaxNode(
-            "fill",
-            {
-                "classifier": "",
-                "seperator": syntax_node.ValueNode("=", str),
-                "data": list_node,
-            },
-        )
+
```
MicahGale, line 348: Wait; was this dead code?

tests/test_universe.py
```diff
@@ -478,3 +466,25 @@ def verify_export(self, fill):
                 assert (old_val == new_val).all()
             else:
                 assert old_val == new_val
+
+
+class TestUniverseGenerators:
```
MicahGale, line 471: I think this should be part of the class `TestUniverse`. I like to organize by the child class being tested more so.

tests/test_universe.py
```diff
@@ -478,3 +466,25 @@ def verify_export(self, fill):
...
+        ],
+    )
+    def test_filled_cells_generator(self, universe, expected_cells):
+        problem = montepy.read_input(
+            os.path.join(self.default_test_input_path, "test_lattice_fill_1.imcnp")
+        )
+        cell_generator = problem.universes[universe].filled_cells
+        filled_cells = [cell.number for cell in cell_generator]
+
+        assert filled_cells == expected_cells
```
MicahGale, line 490: This looks good for testing this function, and you do cover `fill.universe` and `fill.universes`.

MicahGale (changes requested):
Mostly minor tweaks to make. Overall this is a pretty solid PR, and thanks for going above again and helping migrate our test suite.

Side note: 

1. I am trying to get funding to go to PHYSOR 26 next year in Italy. If you are able to attend as well it would be great to meet up.
2. I have some vinyl MontePy laptop stickers that I could mail you for your contributions. If you're interested send me an email.

tests/test_universe.py
MicahGale: Should also test the case where fill is detached from a problem. Easiest way might be to create a new object by montepy.Universe(1).

tests/test_universe.py
kordusas: Seems like it the created value or list_node was not used later.  only self.simple_fill was used. I assumed at some point simple_fill was created on the spot and there was some leftover code. I did not find what it does. Please correct me if I am wrong.

tests/test_universe.py
kordusas: from pytest: 

> Similarly, the following methods are called around each method invocation
> def setup_method(self, method):
>     """ setup any state tied to the execution of the given method in a
>     class.  setup_method is invoked for every test method of a class.
>     """

according to my understanding method is not necessary in this case as we do not have different test cases (yet). I should probably remove the deepcoies then as they become redundant ( `tree = copy.deepcopy(self.tree)`)

doc/source/changelog.rst
```diff
@@ -5,6 +5,15 @@ MontePy Changelog
...
+--------------
+
+**Features Added**
+
+* Added ``Universe.filled_cells``, a generator that yields the cells filled with that universe instance (:issue:`361`).
+* Added ``__eq__`` dunder method to ``Universe`` to support equality comparisons (:issue:`361`).
+* Added a new input example to test the ``Universe.filled_cells`` generator and equality feature (:issue:`361`).
```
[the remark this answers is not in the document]
kordusas: should i just mention that the generator was created, thats enough?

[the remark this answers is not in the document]
MicahGale: I specifically was talking about line 15. I think the other two are good for the changelog.

montepy/universe.py
```diff
@@ -50,7 +50,7 @@ def cells(self):
                     yield cell
 
     @property
-    def filled_cells(self) -> Generator['montepy.Cell', None, None]:
+    def filled_cells(self) -> Generator["montepy.Cell", None, None]:
```
MicahGale, line 53: I'm assuming this is to avoid circular imports? Instead of making it a string import:

``` python
from __future__ import annotations
```
[annotations](https://peps.python.org/pep-0563/)

MicahGale: Also this never yields `None`. Rather if there's nothing to yield it just raises `StopIterator` (or something along those lines).

I double checked this was the case:


``` python
In [11]: uni = montepy.Universe(100)

In [12]: list(uni.filled_cells)
Out[12]: []

In [13]: for cell in uni.filled_cells:
    ...:     print(cell)
    ...:

```

montepy/universe.py
```diff
@@ -50,22 +50,25 @@ def cells(self):
...
+            yield from []
+            return
+
+        for cell in self._problem.cells:
+            if cell.fill:
+                if cell.fill.universes is not None:
+                    if np.any(cell.fill.universes.flatten() == self):
                         yield cell
+                elif cell.fill.universe == self:
+                    yield cell
```
MicahGale, line 71: Fantastic.

montepy/universe.py
```diff
@@ -109,8 +112,3 @@ def __repr__(self):
             f"Problem: {'set' if self._problem else 'not set'}, "
             f"Cells: {[cell.number for cell in self.cells] if self._problem else ''}"
         )
-
-    def __eq__(self, other):
-        if not isinstance(other, type(self)):
-            return False
-        return self.number == other.number
```
MicahGale: I don't think this is what we want to do. I think there might be separate bug in the doctest, and I'm digging into it now.

MicahGale (changes requested):
Getting so close! I don't like removing `__eq__` so I'm going to look into that test failure and see what's going on.

montepy/universe.py
MicahGale: I think the doctest failure is just due to get the problem into a really weird state that I don't understand. Let's just flush it and start over.

On line 1223 of `starting.rst` add a `problem = montepy.read_input("tests/inputs/test.imcnp")` to start fresh for the universe tests. Make sure there's a blank line after `.. testcode::`. Also you will need to remove `4` from the expected output on line `1233` because that's from a mangled start that there was a cell 4.

montepy/universe.py
kordusas: > I'm assuming this is to avoid circular imports? Instead of making it a string import:
> 
> ```python
> from __future__ import annotations
> ```
> 
> [annotations](https://peps.python.org/pep-0563/)

Yes, other suggestion? 


> Also this never yields `None`. Rather if there's nothing to yield it just raises `StopIterator` (or something along those lines).
> 
> I double checked this was the case:
> 
> ```python
> In [11]: uni = montepy.Universe(100)
> 
> In [12]: list(uni.filled_cells)
> Out[12]: []
> 
> In [13]: for cell in uni.filled_cells:
>     ...:     print(cell)
>     ...:
> ```


Yeah, I think it never yields None, so the Generator[YieldType, SendType, ReturnType] type annotation seems accurate.

montepy/universe.py
kordusas: I also like flat structure

montepy/universe.py
kordusas: sure, I wasnt sure which path is preferred, thanks for the answer!

montepy/universe.py
MicahGale: Oh. I never read the docs.

tests/inputs/test_lattice_fill_1.imcnp
kordusas: Ok, I think i will use the test_universe.imcnp I dont see any added value at this point of my input to testing of universe object.

MicahGale (approved):
LGTM. I did confirm that `tests/inputs/test_universe.imcnp` does test both `cell.fill.universe` and `cell.fill.universes` in this case.

Thank you for the contribution!

Issue #103, opened 2022-08-26
Idea: Formatter class


MicahGale:
In GitLab by @tjlaboss on Aug 26, 2022, 13:02

Several Python libraries, such as `logging`, allow users to pass `Formatter` objects.

This could be useful for controlling MontePy outputs. Attributes might be:

 - `.mcnp_version = tuple` (or `.max_line_len = int`); controls maximum line length
 - `.importance_in_data_cards = bool`
 - `.volume_in_data_cards = bool`
 - `.universe_in_data_cards = bool`
 - `.compress_repeat_values = bool`
 - `.float_fmt = str` (ex.: "{:9.6e}")
 - `.line_wrap = enum`; could be "&" or else an "indent" between 5 and some reasonable maximum. 
 - `.format_all_cards = bool`; forces all cards to be mutated.
 - etc.

The syntax could be either `MCNP_Problem.formatter = Formatter(...)` or `MCNP_Problem.write_to_file(..., Formatter(...))`. Basically, MontePy would internally pass around a `Formatter` instead of `mcnp_version`.

PR #757, opened 2025-05-09
Fixed bug with bad rounding for floating point value nodes on export

### LINKED ISSUE #756: Relative Roundoff is Too Large
## Bug Description

When writing new floating point numbers, MontePy reverse-engineers the formatting string in order to preserve formatting. It does not appropriately consider how much of a relative change this is. 

## To Reproduce

 

``` python
import montepy
surf = montepy.Surface("1 SO 3.0")
surf._surface_constants[0].value = 1.249
print(surf.format_for_mcnp_input((6,3,0))[0])
```

Results in 

```
1 SO 1.2
```

MontePy should consider the relative difference that the user is willing to tolerate. There should be a small default number; `eps=1e-5` is probably too large. MontePy should check whether the rounded number varies from the true number by more than `eps`. If so, it should find the maximum amount of rounding that will satisfy the threshold (i.e., determine the number of digits such that `true*(1 - eps) >= rounded >= true*(1 + eps)`.

## Describe alternatives you've considered

Alternatively, write the unrounded true number raising a LineExpansionWarning. 

There are also a few ways to make the threshold user-configurable: 

 1. Global constant that the user can change.
 2. Argument to all `write()` and `format()` functions
 3. Setting in Formatter class: see #103 

I think option 1 should be implemented immediately as a stopgap and option 3 long-term.

Another option could be an absolute threshold.

## Version

 - I suspect 0.2.0 but haven't verified

## Additional context

This is a critical issue.

MicahGale:

### Description

This fixes the bug with MontePy rounding out changes. This bug came from MontePy trying to match the user's original intent.

This was solved by adding a `rel_eps` and `abs_eps` to the formatter which defaults to `1e-6` and `1e-9` respectively. Making this user settable is a later task: #103. The general approach used is:

1. round to the number of digits from `precision`
2. Compare to the user-set value using `math.isclose` using `rel_eps` and `abs_eps`.
3. If they are not close add another digit of precision and repeat.

I tried to think of away that was not iterative using `log` and `exp` but got stuck. It's really easy to tell the most significant digit, but rather hard to find the least significant digit from my quick research. 

Fixes #756 

---

### General Checklist

---

<details open> 

</details>

---

<details>

</details>

---

Ensure that:

----
📚 Documentation preview 📚: https://montepy--757.org.readthedocs.build/en/757/

### DIFF
--- a/montepy/input_parser/syntax_node.py
+++ b/montepy/input_parser/syntax_node.py
@@ -968,8 +968,6 @@ class ValueNode(SyntaxNodeBase):
             "as_int": False,
             "int_tolerance": 1e-6,
             "is_scientific": True,
-            "rel_eps": 1e-6,
-            "abs_eps": 1e-9,
         },
         int: {"value_length": 0, "zero_padding": 0, "sign": "-"},
         str: {"value_length": 0},
@@ -1259,23 +1257,6 @@ class ValueNode(SyntaxNodeBase):
             )
         return self.value != self._og_value
 
-    def _avoid_rounding_truncation(self):
-        if self._formatter["is_scientific"]:
-            pass
-        else:
-            precision = self._formatter["precision"]
-            val = self.value
-        while True:
-            if not math.isclose(
-                val,
-                round(val, precision),
-                rel_tol=self._formatter["rel_eps"],
-                abs_tol=self._formatter["abs_eps"],
-            ):
-                precision += 1
-            break
-        self._formatter["precision"] = precision
-
     def format(self):
         if not self._value_changed:
             return f"{self._token}{self.padding.format() if self.padding else ''}"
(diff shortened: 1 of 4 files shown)


montepy/input_parser/syntax_node.py
```diff
@@ -1257,6 +1259,35 @@ def _value_changed(self):
...
+        while True:
+            if not math.isclose(
+                val,
+                round(val, precision),
+                rel_tol=self._formatter["rel_eps"],
+                abs_tol=self._formatter["abs_eps"],
+            ):
+                precision += 1
+            else:
+                break
```
tjlaboss, line 1288: Simplify this to 

```python
    while not math.isclose(
        val,
        round(val, precision),
        rel_tol=self._formatter["rel_eps"],
        abs_tol=self._formatter["abs_eps"],
    ):
        precision += 1
```

tjlaboss (commented):
I'll do my final review after the merge conflict, but this appears to be working correctly so far.

At some point, we need to include an example and document the default constants.

montepy/input_parser/syntax_node.py
MicahGale: Wow. Of course. SMH.

tjlaboss (approved):

PR #867, opened 2025-12-17
Implement _collection_ref to link objects to their NumberedObjectCollection parent

### LINKED ISSUE #849: Link objects to their NumberedObjectCollection Parent
## Is your feature request related to a problem? Please describe.

This came from a PR's discussion: https://github.com/idaholab/MontePy/pull/841#pullrequestreview-3385642879. Rather than relying on `self._problem.cells` for a cell to update the number cache in `cells` we should make a `self._collection`. This has the benefit of being clear, and more reliable. There is no feasible situation where an object is not attached to a collection, but needs to update a collection's number cache. 

a `weakref` ala `self._problem` is created. Whenever an object is appended (via `__internal_append`?) this `weakref` is attached. 

There is one case where we can't rely on the number cache: secondary collections, e.g., `Cell.surfaces`. In `jit_parsing` I am contemplating reworking this. Perhaps making these "shallow" pointers that just know which objects it owns, and relies on the primary collection for numbered look-ups, or just does an $\mathcal{O}(N)$ search. 

## Describe alternatives you've considered

1. Status quo with the $\mathcal{O}(N^2)$ penalty in edge cases.

digvijay-y:

### Description

Added a ```_collection_ref ``` as a weak_ref which points the object back to its collection. 

# Problems
```deepcopy``` got messed up link "number already in use" another failed test due to surface conflicts. 
`_cell` pointed to wrong cell objects.
 
# Solution:
 Rebuild cache after `deepcopy` instead of overwriting `_collection_ref`.
 Added `_set_cell method` to `halfspace`

```
    100 objects -> 0.127 seconds
    500 objects -> 0.443 seconds
   1000 objects -> 0.937 seconds
   8000 objects -> 8.385 seconds
```

<details>

  ```Py
import time
import montepy


class Collection(montepy.Cells):
    """Use the real Cells collection for testing"""

    pass


def dump_stats(n):
    collection = Collection()
    start = time.time()
    for i in range(n):
        cell = montepy.Cell(f"{i+1} 0 -1")
        collection.append_renumber(cell)
    end = time.time()
    print(f"{n:>7d} objects -> {end - start:.3f} seconds")


if __name__ == "__main__":
    print("Performance test: append_renumber with _collection weakrefs")
    print("=" * 60)
    for n_cells in [100, 500, 1000, 8000]:
        dump_stats(n_cells)
    print("=" * 60)

  ```

</details>

Fixes #849 

---

### General Checklist

- [X] I have performed a self-review of my own code.
- [X] The code follows the standards outlined in the [development documentation](https://www.montepy.org/en/stable/dev_standards.html).
- [X] I have formatted my code with `black` version 25.
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable).

---

### LLM Disclosure

Were any large language models (LLM or "AI") used in to generate any of this code?

- [X] Yes
    - Model(s) used:
    -  Copilot
- [ ] No

<details open> 

- [ ] I have documented all added classes and methods.

</details>

---

<details>

</details>

---

Ensure that:

- [ ] This PR fully addresses and resolves the referenced issue(s).
- [ ] The submitted code is consistent with the merge checklist outlined [here](https://www.montepy.org/en/stable/dev_checklist.html#merge-checklist).

----
📚 Documentation preview 📚: https://montepy--867.org.readthedocs.build/en/867/

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -17,7 +17,6 @@ import sys
 sys.path.insert(0, os.path.abspath("../.."))
 import montepy
 
-
 # -- Project information -----------------------------------------------------
 
 project = "MontePy"
--- a/tests/test_edge_cases.py
+++ b/tests/test_edge_cases.py
@@ -120,9 +120,7 @@ def test_geometry_comments():
       :(-11536    97  -401 )  $ C 3 Lower water
       :(-11546    97  -401 )  $ C 4 Lower water
       :(-11556    97  -401 )  $ C 5 Lower water
-      :(-11576    97  -401 ) imp:n=1 $ C 7 Lower water""".split(
-        "\n"
-    )
+      :(-11576    97  -401 ) imp:n=1 $ C 7 Lower water""".split("\n")
     input_obj = montepy.input_parser.mcnp_input.Input(
         in_strs, montepy.input_parser.block_type.BlockType.CELL
     )
--- a/tests/test_material.py
+++ b/tests/test_material.py
@@ -47,8 +47,7 @@ class TestMaterial:
 
     @pytest.fixture
     def parsed_material(_):
-        return Material(
-            """M1   1001.00c   0.05
+        return Material("""M1   1001.00c   0.05
       1001.04c   0.05
       1001.80c   0.05
       1001.04p   0.05
@@ -64,8 +63,7 @@ class TestMaterial:
       95242      0.05
       27560.50c  0.05
       94239      0.05
-      28000.60c  0.05"""
-        )
+      28000.60c  0.05""")
 
     @pytest.fixture
     def materials(_, big_material, parsed_material):

### CODE REVIEW

File: montepy/mcnp_object.py [hunk 1]
```diff
@@ -107,6 +107,7 @@ def __init__(
         except AttributeError:
             self._BLOCK_TYPE = montepy.input_parser.block_type.BlockType.DATA
         self._problem_ref = None
+        self._collection_ref = None
```
MicahGale: Good design. This should be for `NumberedMCNP_Object` though and not `mcnp_object`. The `_collection` reference only makes sense for numbered objects (for now).

montepy/numbered_mcnp_object.py
```diff
@@ -14,22 +15,31 @@
...
+                # Finding via inheritance
+                for obj_class in obj_map:
+                    if isinstance(self, obj_class):
+                        collection_type = obj_map[obj_class]
+                        break
+
+            if collection_type is not None:
+                collection = getattr(self._problem, collection_type.__name__.lower())
+            else:
+                collection = None
```
MicahGale: This should still raise an error eventually.

montepy/numbered_object_collection.py
```diff
@@ -196,8 +197,18 @@ def link_to_problem(self, problem):
             self._problem_ref = None
         else:
             self._problem_ref = weakref.ref(problem)
+        # Rebuild the number cache to ensure it reflects current object numbers
+        # after deepcopy/unpickling when the cache might be stale
+        self.__num_cache.clear()
+        for obj in self._objects:
+            self.__num_cache[obj.number] = obj
```
MicahGale: I'm not positive this is necessary. I thought it was handled through the MCNP_problem side.

montepy/numbered_object_collection.py
```diff
@@ -196,8 +197,18 @@ def link_to_problem(self, problem):
...
         for obj in self:
             obj.link_to_problem(problem)
+            # the _collection_ref that points to the main cells collection.
+            if problem is not None:
+                existing_coll = obj._collection
+                if existing_coll is None or existing_coll._problem is not problem:
+                    obj._collection_ref = weakref.ref(self)
```
MicahGale: You should make a `_link_to_collection` method instead of doing this manually every time.

MicahGale (changes requested):
I didn't fully evaluate how this works, but I think there are some design changes needed first.

digvijay-y:
1. Moved `_collection_ref` to `Numbered_MCNP_Object`
2. Removed _set_cell methods
3. Fixed `_number_validator` to raise error
4.  Created _link_to_collection method
5. Removed cache rebuild
6.  Fixed try block error message

montepy/cell.py
```diff
@@ -927,6 +927,10 @@ def num(obj):
         for key in keys:
             attr = getattr(self, key)
             setattr(result, key, copy.deepcopy(attr, memo))
+        # Clear weakrefs so the cloned cell isn't linked to original collection/problem
+        # This prevents number conflict checks against the original collection
+        result._collection_ref = None
+        result._problem_ref = None
```
MicahGale: This is still not the desired behavior in all cases. Generally they should be added to the same problem. I know that this can lead to a number collision in an edge case with hypothesis, that I haven't fixed yet, and I just ignore it with `rm -r .hypothesis`. I think that's a separate issue from this, and I need to open a bug report. Try it without these lines, and with purging the `.hypothesis` folder.

montepy/numbered_mcnp_object.py
```diff
@@ -14,20 +15,30 @@
...
-                if isinstance(self, obj_class):
-                    collection_type = obj_map[obj_class]
-                    found = True
-                    break
-            if not found:
-                raise e
-        collection = getattr(self._problem, collection_type.__name__.lower())
+
+    # Only validate against collection if linked to a problem
+    if self._problem is not None:
```
MicahGale: You can skip a level of `if`s and just check `if self._collection is not None`.

montepy/numbered_mcnp_object.py
```diff
@@ -14,20 +15,30 @@
...
+                        collection_type = obj_map[obj_class]
+                        break
+
+            if collection_type is not None:
+                collection = getattr(self._problem, collection_type.__name__.lower())
+            else:
+                raise TypeError(
+                    f"Could not find collection type for {type(self).__name__} in problem."
+                )
+
```
MicahGale: I think you can delete this whole branch. I don't see a case where `self._problem` is linked, but `self._collection` is not.

montepy/numbered_mcnp_object.py
```diff
@@ -123,6 +135,24 @@ def _add_children_objs(self, problem):
                 except (TypeError, AssertionError):
                     prob_collect.append(child_collect)
 
+    @property
+    def _collection(self):
+        """Returns the parent collection this object belongs to, if any."""
```
MicahGale: You need to also reimplement `link_to_problem` here. It should at least:

1. call `super().link_to_problem`
2. grab the relevant collection from `problem`. See the logic above in `_validate_number`. 
3. create a `weakref.ref` to that collection (in `_collection_ref`). 

This will be the secondary way to update this ref, besides appending.

MicahGale: Though this may lead to weird situations where objects are linked to problems without being in the collections. I'm going to belay this until I read further.

montepy/numbered_object_collection.py
```diff
@@ -198,13 +199,38 @@ def link_to_problem(self, problem):
...
+                if existing_coll is None or existing_coll._problem is not problem:
+                    self._link_to_collection(obj)
 
     @property
     def _problem(self):
         if self._problem_ref is not None:
             return self._problem_ref()
         return None
 
+    def _link_to_collection(self, obj):
```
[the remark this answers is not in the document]
MicahGale: The reason being is objects should respect each other's private attributes. I am ok with calling each other's private functions as that in my mind is more marking it as internal use only.

MicahGale (changes requested):
This is a big improvement, and much more like what I was thinking. I have some more feedback, but it is not major.

Looking at the data you collected this does look like it is indeed linear (O(N)), so that's great that it is no longer O(N^2). I have been thinking more about how to present performance data recently. I think in general, rates should generally be used (e.g., objects/second) over just raw time. My last paper on this though did not use rates, so I'm working on this as well.

digvijay-y:
I can try data in object/second

MicahGale (changes requested):
A few of my old conversations were resolved a bit early, and still need to be addressed. Also one line of code is no longer covered by the test suite.

digvijay-y:
@MicahGale I guess there's some issue with CI, I tried Black . multiple times, but CI doesn't approve it.

PR #915, opened 2026-02-28
Remove Guardrail from NumberedObjectCollection

### LINKED ISSUE #895: Remove Guardrails from NumberedObjectCollection
Originally `NumberedObjectCollection` was very wary of `self.__num_cache` because it was using "mutable data" as the keys. With #867 now though there should in theory be no case in which the cache is not up to date. Right now whenever this is a cache miss it is assumed it is a true miss, and not that the object is missing, so an $O(N)$ search is completed before a `KeyError` is raised. This is expensive, especially for renumbering tasks. 

I believe now that this guardrail of search before a `KeyError` is no longer necessary. [See this guardrail that is no longer covered because of this](https://coveralls.io/builds/77662955/source?filename=montepy%2Fnumbered_object_collection.py#L1097).

digvijay-y:

### Description

Removed Guardrail from `NumberedObjectCollection` 
 [See this guardrail that is no longer covered because of this](https://coveralls.io/builds/77662955/source?filename=montepy%2Fnumbered_object_collection.py#L1097). 
 
 - Updated `check_number()` in `NumberObjectCollection` to follow changes
 
The problem is that test intent is already covered by the fact that `clone_region=True` clones the surface and assigns it a brand new number anyway. So by the time `verify_clone_format` is called, surf.number is already a freshly assigned number, not 2 anymore. The temporary mutation to 1000 was redundant and just added fragility.

 - Updated `verify_clone_format` test in `test_cell_problem` 

Fixes #895

---

### General Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable).

---

### LLM Disclosure

1. Are you?

   - [x] A human user 

  - [ ] Yes
      - Model(s) used:
  - [x] No

<details open> 

- [ ] I have documented all added classes and methods.

</details>

---

<details>

</details>

---

Ensure that:

- [x] This PR fully addresses and resolves the referenced issue(s).
- [x] The submitted code is consistent with the merge checklist outlined [here](https://www.montepy.org/en/stable/dev_checklist.html#merge-checklist).
- [x] The PR covers all relevant aspects according to the development guidelines.
- [x] 100% coverage of the patch is achieved, or justification for a variance is given.

----
📚 Documentation preview 📚: https://montepy--915.org.readthedocs.build/en/915/

### DIFF
--- a/montepy/numbered_object_collection.py
+++ b/montepy/numbered_object_collection.py
@@ -613,7 +613,15 @@ class NumberedObjectCollection(ABC):
             raise TypeError(f"object being appended must be of type: {self._obj_class}")
         if not isinstance(step, Integral):
             raise TypeError("The step number must be an int")
-        number = obj.number if obj.number > 0 else 1
+        # If the object has no number yet (None) or an invalid number (<=0),
+        # start renumbering from 1.  We must assign a concrete number before
+        # calling append() because __internal_append requires obj.number to be
+        # a non-None integer.
+        if obj.number is None or obj.number <= 0:
+            number = 1
+            obj.number = number
+        else:
+            number = obj.number
         if self._problem:
             obj.link_to_problem(self._problem)
         obj._unlink_from_collection()
--- a/montepy/numbered_object_collection.py
+++ b/montepy/numbered_object_collection.py
@@ -613,15 +613,9 @@ class NumberedObjectCollection(ABC):
             raise TypeError(f"object being appended must be of type: {self._obj_class}")
         if not isinstance(step, Integral):
             raise TypeError("The step number must be an int")
-        # If the object has no number yet (None) or an invalid number (<=0),
-        # start renumbering from 1.  We must assign a concrete number before
-        # calling append() because __internal_append requires obj.number to be
-        # a non-None integer.
         if obj.number is None or obj.number <= 0:
-            number = 1
-            obj.number = number
-        else:
-            number = obj.number
+            obj.number = 1
+        number = obj.number
         if self._problem:
             obj.link_to_problem(self._problem)
         obj._unlink_from_collection()
--- a/montepy/numbered_object_collection.py
+++ b/montepy/numbered_object_collection.py
@@ -254,12 +254,9 @@ class NumberedObjectCollection(ABC):
             raise ValueError(f"The number must be non-negative. {number} given.")
         conflict = False
         # only can trust cache if being updated
-        if self._problem:
-            if number in self.__num_cache:
-                conflict = True
-        else:
-            if number in self.numbers:
-                conflict = True
+        if number in self.__num_cache:
+            conflict = True
+
         if conflict:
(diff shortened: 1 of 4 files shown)


montepy/numbered_object_collection.py
```diff
@@ -1125,17 +1122,9 @@ def get(self, i: int, default=None) -> (Numbered_MCNP_Object, None):
...
-            pass
-        for obj in self._objects:
-            self.__num_cache[obj.number] = obj
-            if obj.number == i:
-                self.__num_cache[i] = obj
-                return obj
+        ret = self.__num_cache.get(i)
+        if ret is not None and ret.number == i:
+            return ret
         return default
```
MicahGale: This could be further simplified to be just:

``` suggestion
        return self.__num_cache.get(i, default)
```

tests/test_cell_problem.py
```diff
@@ -275,9 +275,6 @@ def verify_internal_links(cell):
 
 def verify_clone_format(cell):
     surf = list(cell.surfaces)[0]
-    old_num = surf.number
-    num = 1000
-    surf.number = num
```
MicahGale, line 280: I'll have to think if this is the best way to update the test.

MicahGale (changes requested):
Some code simplification, and I want to think more about the test changes. 

Side note: I am trying out having Copilot do a first review. Does it seem helpful to you @digvijay-y?

digvijay-y:
> I am trying out having Copilot do a first review. Does it seem helpful to you @digvijay-y?

It helps. But let's test it with large changes.

MicahGale:
> It helps. But let's test it with large changes.

I think it would also be a lot more helpful if I gave it instructions from the contributing guide, and developer documentation, etc.

tests/test_cell_problem.py
digvijay-y: @MicahGale  got any suggestions here, or should I revert it back and find a way......

MicahGale: No I think it's good to keep out. Not sure if this is causing the tests to fail all of a sudden though. Need any help debugging?

MicahGale:
I can look into this this weekend if you would like.

digvijay-y:
> I can look into this this weekend if you would like.

Yes, please. I’ve been tied up with a few other tasks.... I will push some changes.

digvijay-y:
I have removed the `force num_cache miss` block. It does pass the case, as with guardrail removed, it won't be the same. I'm not sure if its the right approach...

MicahGale:
> I have removed the `force num_cache miss` block. It does pass the case, as with guardrail removed, it won't be the same. I'm not sure if its the right approach...

I'll have to review it again, but I think it should be unnecessary now as cache misses should be impossible.

MicahGale (approved):
LGTM. Thanks @digvijay-y

PR #929, opened 2026-03-14
fix(surface): preserve original case of surface type mnemonics

### LINKED ISSUE #522: Preserve case when writing surface cards
Presently, MontePy converts surface cards to UPPERCASE when writing out an input.

Preserve the original case.

**MCNP input file snippet**

original:
```
Case-sensitive example
C Cells
1   1 -1.0    -1  
2   1 -0.5    -2 +1
3   0            +2

c Surfaces get changed
1    sO 0
2    So 2

c Data
m1      8016.71c  2-02 $ comment 1
        8017.71c  9-09 $ comment 2
imp:p   2 1 0 
mode p
nps 1
```

after writing:
```
Case-sensitive example
C Cells
1   1 -1.0    -1  
2   1 -0.5    -2 +1
3   0            +2

c Surfaces get changed
1    SO 0
2    SO 2

c Data
m1      8016.71c  2-02 $ comment 1
        8017.71c  9-09 $ comment 2
imp:p   2 1 0 
mode p
nps 1
```

novavale:
### Description

Fixes #522

MCNP surface type mnemonics are case-insensitive, so users may write `sO`, `So`, or `so` instead of `SO`. Previously MontePy always wrote them back in uppercase, silently changing the user's input.

**Root cause:** `ValueNode.convert_to_enum(switch_to_upper=True)` did not set `_og_value` after resolving the enum. As a result, `_value_changed` always returned `True` (because `_og_value` was `None`), causing `format()` to fall through to the enum-formatting path — which emits `value.value` (always uppercase, e.g. `"SO"`) instead of returning the original token verbatim.

**Fix:** one line in `convert_to_enum` — set `self._og_value = self._value` after the enum is resolved. This mirrors the identical pattern in `convert_to_int` (line 1011). When the user hasn't changed the surface type, `_value_changed` is `False` and `format()` returns the original token, preserving case.

Before:
```mcnp
1    sO 0     → written as: 1    SO 0
```

After:
```mcnp
1    sO 0     → written as: 1    sO 0   ✓
```

---

### General Checklist

---

### LLM Disclosure

1. Are you?

   - [ ] A human user
   - [x] A large language model (LLM), including ones acting on behalf of a human

  - [x] Yes
      - Model(s) used: Claude (claude-sonnet-4-6)
  - [ ] No

<details open>

</details>

---

- Change is a single line in `syntax_node.py:convert_to_enum` — minimal surface area
- 540 tests pass; 1 pre-existing failure in `test_cells_clone` (known hypothesis flakiness, unrelated to this change)
- The fix applies to all enum conversions via `convert_to_enum`, which in MontePy is called from `surface.py` (surface type) and `lattice_input.py` (lattice type, integer-based). Both benefit from the same case/value preservation logic.

----
📚 Documentation preview 📚: https://montepy--929.org.readthedocs.build/en/929/

### DIFF
--- a/montepy/input_parser/syntax_node.py
+++ b/montepy/input_parser/syntax_node.py
@@ -1056,7 +1056,6 @@ class ValueNode(SyntaxNodeBase):
         if not (allow_none and self._value is None):
             self._value = enum_class(value)
         self._formatter = self._FORMATTERS[format_type].copy()
-        self._og_value = self._value
 
     def convert_to_str(self):
         """Converts this ValueNode to being a string type.
@@ -1258,6 +1257,17 @@ class ValueNode(SyntaxNodeBase):
             return not math.isclose(
                 self._print_value, self._og_value, rel_tol=rel_tol, abs_tol=abs_tol
             )
+        if isinstance(self._type, type) and issubclass(self._type, enum.Enum):
+            # Compare case-insensitively: the original token may differ in case
+            # from the normalised enum value (e.g. "sO" → SurfaceType.SO).
+            # If the canonical value matches the original token modulo case,
+            # the user has not changed the field and the original token is kept.
+            og = (
+                self._og_value
+                if isinstance(self._og_value, str)
+                else str(self._og_value)
+            )
+            return self.value.value.upper() != og.upper()
         return self.value != self._og_value
 
     def _avoid_rounding_truncation(self):
--- a/montepy/input_parser/syntax_node.py
+++ b/montepy/input_parser/syntax_node.py
@@ -1258,16 +1258,25 @@ class ValueNode(SyntaxNodeBase):
                 self._print_value, self._og_value, rel_tol=rel_tol, abs_tol=abs_tol
             )
         if isinstance(self._type, type) and issubclass(self._type, enum.Enum):
-            # Compare case-insensitively: the original token may differ in case
-            # from the normalised enum value (e.g. "sO" → SurfaceType.SO).
-            # If the canonical value matches the original token modulo case,
-            # the user has not changed the field and the original token is kept.
-            og = (
-                self._og_value
-                if isinstance(self._og_value, str)
-                else str(self._og_value)
-            )
-            return self.value.value.upper() != og.upper()
+            canonical = self.value.value
+            if isinstance(canonical, str):
+                # Compare case-insensitively: the original token may differ in
+                # case from the normalised enum value (e.g. "sO" → SO).
+                # If they match modulo case the field is unchanged and the
+                # original token casing is preserved on write.
+                #
+                # _og_value may not be a string for integer-based enums (e.g.
+                # Lattice), where convert_to_int() runs before convert_to_enum()
+                # leaving _og_value as an int.  Normalise to str for the
+                # case-insensitive compare.
+                og = (
+                    self._og_value
+                    if isinstance(self._og_value, str)
+                    else str(self._og_value)
+                )
+                return canonical.upper() != og.upper()
+            # Integer-valued enums (e.g. Lattice): case is irrelevant; fall
+            # through to the exact-equality check below.
         return self.value != self._og_value
 
     def _avoid_rounding_truncation(self):
--- a/montepy/input_parser/syntax_node.py
+++ b/montepy/input_parser/syntax_node.py
@@ -1056,6 +1056,7 @@ class ValueNode(SyntaxNodeBase):
         if not (allow_none and self._value is None):
             self._value = enum_class(value)
         self._formatter = self._FORMATTERS[format_type].copy()
+        self._og_value = self._value
 
     def convert_to_str(self):
         """Converts this ValueNode to being a string type.
--- a/tests/test_surfaces.py
+++ b/tests/test_surfaces.py
@@ -603,3 +603,22 @@ def test_surface_is_white_bound_setter(cp_simple_problem, in_str, expected):
     verify_prob_export(cp_simple_problem, surf)
     with pytest.raises(TypeError):
         surf.is_white_boundary = 1
+
+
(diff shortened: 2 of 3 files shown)


montepy/input_parser/syntax_node.py
```diff
@@ -1056,6 +1056,7 @@ def convert_to_enum(
         if not (allow_none and self._value is None):
             self._value = enum_class(value)
         self._formatter = self._FORMATTERS[format_type].copy()
+        self._og_value = self._value
```
MicahGale: Shouldn't this have been set when the `ValueNode` is first initialized?

tests/test_surfaces.py
```diff
@@ -603,3 +603,22 @@ def test_surface_is_white_bound_setter(cp_simple_problem, in_str, expected):
...
+        ("1 SO 0.0", "1 SO 0.0"),
+        ("1 pz 0.0", "1 pz 0.0"),
+        ("1 PZ 0.0", "1 PZ 0.0"),
+        ("1 Pz 0.0", "1 Pz 0.0"),
+    ],
+)
+def test_surface_preserves_mnemonic_case(in_str, expected):
+    """Surface mnemonics should be written with the original case (issue #522)."""
+    surf = Surface(in_str)
+    result = surf.format_for_mcnp_input((6, 2, 0))[0]
```
MicahGale: Use `surf.mcnp_str()`

MicahGale (changes requested):
A few questions. Also the failing tests need to be resolved first.

novavale:
Thanks for the review — addressed in 4a101875.

**"Shouldn't this have been set when the `ValueNode` is first initialized?"**

You're right. `__init__` already sets `_og_value = self.value` (the raw string token, e.g. `"sO"`), so there's no need to overwrite it in `convert_to_enum`. I reverted that change.

Instead, I extended `_value_changed` to handle enum types: when `_type` is an `enum.Enum` subclass, the comparison is done case-insensitively between `self.value.value` (the canonical enum string, e.g. `"SO"`) and `self._og_value` (the original token, e.g. `"sO"`). If they match modulo case, the field is considered unchanged and `format()` returns the original token verbatim. If the user changes the surface type (e.g. to `SurfaceType.PZ`), `_value_changed` returns `True` as expected.

**"Use `surf.mcnp_str()`"**

Updated.

**Failing tests**

`changelog-test` — added a changelog entry under `#Next Version# / Bugs Fixed`.

`doc-build` — the failure is in "Test for broken hyperlinks", which is pre-existing on `develop` itself (confirmed: the most recent develop CI run also fails this step). It's unrelated to this PR.

montepy/input_parser/syntax_node.py
```diff
@@ -1257,6 +1257,17 @@ def _value_changed(self):
...
+        if isinstance(self._type, type) and issubclass(self._type, enum.Enum):
+            # Compare case-insensitively: the original token may differ in case
+            # from the normalised enum value (e.g. "sO" → SurfaceType.SO).
+            # If the canonical value matches the original token modulo case,
+            # the user has not changed the field and the original token is kept.
+            og = (
+                self._og_value
+                if isinstance(self._og_value, str)
+                else str(self._og_value)
+            )
```
MicahGale: What cases are you guarding against for `_og_value` not being a string? `None` was previously checked.

MicahGale:
In the failed tests it appears you can't always assume you are working with a string based `Enum`. I'm not sure if we explicitly use `StrEnum`.

montepy/input_parser/syntax_node.py
novavale: Fixed — that line (`self._og_value = self._value` inside `convert_to_enum`) has been removed. The original value is now captured only at `__init__` time, so `_og_value` correctly holds the raw string token (e.g. `"sO"`) throughout the object's lifetime.

tests/test_surfaces.py
novavale: Done — the test now uses `surf.mcnp_str()` as requested.

montepy/input_parser/syntax_node.py
novavale: Good catch — the CI failure (`AttributeError: 'int' object has no attribute 'upper'`) in `test_expanding_new_line` revealed the case. For integer-valued enums like `Lattice`, `convert_to_int()` is called before `convert_to_enum()`, which leaves `_og_value` as an int. The `None` guard at the top of `_changed()` only handles `None`; it doesn't cover this.

The fix (pushed in the latest commit) restricts the case-insensitive path to string-valued enums only — by checking `isinstance(canonical, str)` first — and falls through to the existing exact-equality comparison for integer-valued enums. The `isinstance(self._og_value, str)` guard is retained (with an explanatory comment) to handle the int-og-value case within the string-canonical branch.

novavale:
Addressed in 5c9ef5ae — added handling for both enum flavours:

- **String-valued enums** (e.g. `SurfaceType`): compare case-insensitively. The `isinstance(self._og_value, str)` guard is needed here because `convert_to_int()` can precede `convert_to_enum()` (as with `Lattice`), leaving `_og_value` as an `int`. When that happens we normalise to `str` before comparing.
- **Integer-valued enums** (e.g. `Lattice`): `canonical = self.value.value` is an `int`, so case comparison doesn't apply; fall through to the existing `!=` check.

Note: `test_cells_clone` in `test_numbered_collection.py` is already failing on `develop` and is unrelated to this PR.

novavale:
Addendum: `tests/test_version.py::test_version` also fails on `develop` with the same assertion error (expects `"Undefined"` but gets the actual version string). Both pre-existing failures reproduce cleanly on `develop` with no changes from this PR.

MicahGale (approved):
LGTM. Thanks @novavale.

Issue #73, opened 2022-05-25
Add translations to surfaces


MicahGale:
In GitLab by @tjlaboss on May 25, 2022, 11:17

Let's say you want to apply x, y, z translations to all the surfaces (or a subset of surfaces) in a problem. Maybe you want to add or remove a transformation card, or maybe you just want to shift things around without transformations.

Various surface types that would need implementations:
 - Those infinite in the axis of shift (no change to cz or c/z for delta-z)
 - Those moving only along the axis of shift (do change to c/y for delta-z)
 - Those that need to change the surface type (change a cy to c/y for delta-z)
 - Macrobodies
 - Those with more complicated quadratic equations (plane equations, general surface)

tjlaboss:
I opened #630, potentially a prerequisite to this issue.

PR #940, opened 2026-03-30
Made a metaclass for building surface classes from specifications to handle almost all surface types. 

### LINKED ISSUE #502: Add support for all surface types
I'm a bit surprised that this issue didn't exist yet, unless I suck at searching.

The goal is to semantically support all surface types ala `CylinderOnAxis` so you can do things like `cylinder.radius =...`.

The goal is to support all of Table 5.1 from the 6.3 manual:
image

Given how similar all of these surfaces are implemented I wonder if we could use meta-programming to have python write these classes.

### LINKED ISSUE #935: Implement Tori
## Is your feature request related to a problem? Please describe.

Tori are very helpful in fusion. So it would be helpful to have these supported prior to #502.

A new class or classes like `TorusOnAxis`. See #877 for basis. 

Instead of `radius` will need to use `minor_radius` and `major_radius`. 

## Additional context

This will be very helpful for an upcoming workshop.

MicahGale:

### Description

This project started as a fun weekend project of me asking copilot for a prototype. After seeing what it made I realized it was actually pretty straight forward, so I made the metaclass myself (mostly). Then I bugged Claude to write all of the specifications. This was partially an experiment in having AI as an unpaid intern. 

This created a MetaClass `_SurfaceClassFactory` that can create a `Surface` class from a minimal specification. The specification is given by `@dataclasses` (I wanted to play around with them) that primarily specifies what each surface parameter/constant means. These specifications are then used to make new properties that are attached to the class (via `namespace`) prior to the class being created, since it is a MetaClass after all. 

In the documentation I gave guidance that users should start using `ZCylinder` instead of `CylinderOnAxis`, but that the latter won't be deprecated. Is this reasonable guidance?

Fixes #502, fixes #935

# Questions:

1. should the one-letter variable parameters, e.g., `z`, `a`, etc., be lower or upper_case?

---

### General Checklist

---

### LLM Disclosure

1. Are you?

   - [x] A human user 

  - [x] Yes
      - Model(s) used: Copilot for prototyping. Claude (Sonnet 4.6) for actual implementation. 
  - [ ] No

<details open> 

- [x] For infrastructure updates, I have updated the developer's guide.
- [x] For significant new features, I have added a section to the getting started guide.

</details>

---

<details>

</details>

---

Ensure that:

- [ ] This PR fully addresses and resolves the referenced issue(s).
- [ ] The submitted code is consistent with the merge checklist outlined [here](https://www.montepy.org/en/stable/dev_checklist.html#merge-checklist).

----
📚 Documentation preview 📚: https://montepy--940.org.readthedocs.build/en/940/


MicahGale:
TODO:

* [x] update the tutorial guide in the documentation
* [x] Check over all the documentation
* [x] Write it up in the developer guide
* [x] Remove references to surface types in limitations.
* [x] Fix coverage.
* [x] Fix CX, CY, CZ
* [x] Fix K/X, K/Y, K/Z
* [x] Fix TX, TY, TZ 
  * [x]  how to handle +/-1?
* [x] link to sign from discussion of nappe on cones.

MicahGale:
@tjlaboss this is ready for review.

montepy/surfaces/axis_plane.py
```diff
@@ -1,80 +1,4 @@
-# Copyright 2024, Battelle Energy Alliance, LLC All Rights Reserved.
+# Copyright 2024-2025, Battelle Energy Alliance, LLC All Rights Reserved.
```
tjlaboss: Copyrights throughout?

montepy/surfaces/axis_plane.py
```diff
@@ -1,80 +1,4 @@
...
-                                        surface.transform, tolerance
-                                    ):
-                                        ret.append(surface)
-                            else:
-                                if surface.transform is None:
-                                    ret.append(surface)
-            return ret
-        else:
-            return []
+__all__ = ["AxisPlane"]
```
tjlaboss: Many surface modules have been replaced with just 2 lines. 

Is there any reason to maintain these separate files? Is that reason only for maintaining reverse compatibility with `montepy.surfaces.axis_plane.AxisPlane` until the next major release?

montepy/surfaces/surface_builder.py
```diff
@@ -1,13 +1,120 @@
...
-from montepy.surfaces.surface import Surface, InitInput
+from montepy.surfaces.surface import (
+    ArbitraryPolyhedron,
+    AxisAlignedQuadric,
+    AxisPlane,
+    Box,
+    ConeOnAxis,
+    ConeParAxis,
+    CylinderOnAxis,
+    CylinderParAxis,
```
tjlaboss: Do we plan to deprecate the old, clunkier names at the next major release?

Or, will these stay used internally?

montepy/surfaces/surface.py
```diff
@@ -22,7 +28,156 @@ def _surf_type_validator(self, surf_type):
...
+    tuple_length: int = 1
+
+
+@dataclass
+class _SurfaceTypeSpec:
+    """
+    Specification for a class for a surface type.
+    """
+
+    surface_types: Set[SurfaceType]
```
tjlaboss: `Set` is not imported

montepy/surfaces/surface.py
```diff
@@ -22,7 +28,156 @@ def _surf_type_validator(self, surf_type):
...
+
+class _SurfaceClassFactory(_ExceptionContextAdder):
+    """A metaclass for building :class:`Surface` instances.
+
+    This will take a specification and add the necessary properties
+    to the class. It will load the required class attributes so that
+    the init function is able to load the right data into the new attributes.
+    """
+
+    def __new__(cls, name, bases, namespace, spec: SurfaceTypeSpec = None, **kwargs):
```
tjlaboss: `SurfaceTypeSpec` is not imported.

montepy/surfaces/surface.py
```diff
@@ -118,28 +277,55 @@ def __init__(
...
         int
         """
-        return 1
+        return cls._NUM_PARAMS
 
-    @staticmethod
-    def _allowed_surface_types():
+    @classmethod
+    def _allowed_surface_types(cls):
         """ "
```
tjlaboss: trailing `"`

tests/test_surfaces.py
```diff
@@ -18,7 +18,45 @@
 from montepy.surfaces.general_sphere import GeneralSphere
 from montepy.surfaces.sphere_at_origin import SphereAtOrigin
 from montepy.surfaces.sphere_on_axis import SphereOnAxis
-from montepy.surfaces.surface import Surface
+from montepy.surfaces.surface import (
```
tjlaboss: 1. See prior remark about just moving all these to `surface.py`
2. Would this be one of those times to break the rules and `import *`?

tjlaboss (changes requested):
Overall a useful PR that replaces a lot of verbose code with some much-needed abstraction. A few nitpicks, a few small questions, and one big question.

How do we want to handle the parameter `surface_type` going forward?

Example: Consider `ZCylinder`. In the current implementation, this is a `CylinderOnAxis` on the $z$-axis: a `CZ` and not a `C/Z`.

```python
>>> ZCylinder(number=1).mcnp_str()
montepy.exceptions.IllegalState: Surface: 1 does not have a surface type set.

>>> ZCylinder(number=1, surface_type="C/Z").mcnp_str()
ValueError: ZCylinder must be a surface of type: ['CZ']
```

It seems that this should do one of the following:
 - Delete the parameter `surface_type`, since it must always be `CZ`.
 - Support `C/Z` as an acceptable `surface_type`
 - Delete the parameter `surface_type`, and set the attribute to `CZ` if no coordinates are specified, or else `C/Z`: like `openmc.ZCylinder`.

That does not have to be addressed in this PR, as the interface change might necessitate a major release. A followon issue should be opened if not addressed here.

tjlaboss:
> Should the one-letter variable parameters, e.g., z, a, etc., be lower or upper_case?

The MCNP manual uses both uppercase and lowercase letters, depending on the variable. An argument could be made that MontePy should follow this convention. All-lowercase letters is less confusing and more Pythonic.

image

montepy/surfaces/axis_plane.py
MicahGale: Yeah.... I don't know what happened there.

MicahGale:
> How do we want to handle the parameter `surface_type` going forward?
> ...

My initial thought is that we restrict the accepted values for `surface_type` to those listed, which is already done via `surf_type_validator`. Then for cases when one type is accepted `PZ` <-> `ZPlane` this is just defaulted to. This would avoid ambiguity and unnecessary `IllegalState`

As for `CZ` v. `C/Z` this is currently two classes: `ZCylinder` and `ZCylinderParAxis`. Right now we have no way to go between the two, which I think is fine for now. In the future it would need to be figured out for #73. 

As for `surface_type` in `__init__` yeah I think we should delete. 

I'm thinking:

1. default to the correct surface type when safe to do so.
2. For these cases load a different `__init__` that drops the `surface_type` arg.

Sound good @tjlaboss?

montepy/surfaces/axis_plane.py
MicahGale: Yep, it's to keep the full old import. Should we do `DeprecationWarnings`?

montepy/surfaces/surface_builder.py
MicahGale: See docs. Currently I like the inheritance actually. So I was thinking: discourage direct use, but keep them in place. With how they are implemented I don't think they are at risk of becoming cruft.

tests/test_surfaces.py
MicahGale: I don't want to import the `_Private` metaclass stuff. But I think that's ignored.

MicahGale: I could have sworn a year ago I got this answer in like a minute. There goes the internet, but someone on SO says I'm right so let's do it!

MicahGale:
> The MCNP manual uses both uppercase and lowercase letters, depending on the variable. An argument could be made that MontePy should follow this convention. All-lowercase letters is less confusing and more Pythonic.

I like the idea of this. The one thought I had was: is using `x` confusing because it's double in use? Like the equation for an x-plane would be: $x-x=0$. In the manual they like bar, I'm not sure I like that. Maybe: `x_0`, (or `x_bar`).

MicahGale:
> I like the idea of this. The one thought I had was: is using `x` confusing because it's double in use? ...

I went ahead with keeping them as `x` in context it is clear enough, and easy to use. I did switch the equations to use `x_0`, etc. though.

tjlaboss:
> Sound good @tjlaboss?

yes.

tests/test_surfaces.py
tjlaboss: Yes, \_underscored objects are ignored from `import *`. Alternatively, a subset of the contents of `surface` could put put into `__all__`.

montepy/surfaces/axis_plane.py
tjlaboss: Yes.

montepy/surfaces/surface_builder.py
tjlaboss: I don't see users ever interacting with these classes instead of the new, sleek surface type classes. Perhaps they should be considered as private or as abstract base classes.

MicahGale: I can see users doing: `if isinstance(surf, montepy.CylinderOnAxis):` to work with all `CX`, `CY`, and `CZ`...

tjlaboss: Fair point

MicahGale:
@tjlaboss I deprecated all of those classes now.

montepy/surfaces/axis_plane.py
MicahGale: All deprecated now.

tjlaboss (approved):

PR #972, opened 2026-06-23
Add `in` operator support to comments

### LINKED ISSUE #185: Implement "in" for Comments
It would be very useful to be able to say `if "foo" in cell.leading_comments` to be able to find a cell by comments and not by number.

Also maybe implement the same thing for all comments including dollar sign ones?

yrasool:
### Description

Lets objects be found by their comments, in two layers:

1. `CommentNode.__contains__`: a comment's text can be searched with the `in` operator. It delegates to the existing `contents` property, so it works for both `c` and `$` comments and across appended comment lines. Non-string operands raise `TypeError`, matching normal `str` containment.

2. `CommentCollection`: a new read-only collection type returned by `MCNP_Object.comments` and `leading_comments`. It is its own `collections.abc.Collection` (not a `list` or `tuple` subclass, per review) and stores its comments in a tuple, so it is immutable and supports no operands. It keeps indexing, slicing, iteration, and `len()`. A string checked with `in` searches the text of all comments. Anything else keeps normal membership behavior. `search(pattern)` takes a substring or a compiled `re.Pattern`:

```python
if "shielding" in cell.comments:
    ...
cell.comments.search(re.compile(r"fuel", re.IGNORECASE))
```

`NumberedObjectCollection.get_by_comment` is left to #982, and `trailing_comment` handling to the design discussion in #185.

Fixes #185

---

### General Checklist

---

### LLM Disclosure

1. Are you?

   - [x] A human user

  - [x] Yes
      - Model(s) used: Claude (Claude Code): Opus 4.8, Fable 5
  - [ ] No

<details open>

</details>

---

<details open>

- [x] If this is your first contribution, add yourself to `pyproject.toml` if you wish to do so.

</details>

Ensure that:

- [x] This PR fully addresses and resolves the referenced issue(s).
- [x] The submitted code is consistent with the merge checklist outlined [here](https://www.montepy.org/en/stable/dev_checklist.html#merge-checklist).
- [x] The PR covers all relevant aspects according to the development guidelines.
- [x] 100% coverage of the patch is achieved, or justification for a variance is given.

----
📚 Documentation preview 📚: https://montepy--972.org.readthedocs.build/en/972/

----
📚 Documentation preview 📚: https://montepy--972.org.readthedocs.build/en/972/

### DIFF
--- a/doc/source/api/modules.rst
+++ b/doc/source/api/modules.rst
@@ -33,6 +33,7 @@ Collections
    :template: myclass.rst
 
    montepy.Cells
+   montepy.CommentCollection
    montepy.Materials
    montepy.Surfaces
    montepy.Transforms
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -12,17 +12,18 @@ Next Release
 **Features Added**
 
 * Added ``HalfSpace.replace`` to swap dividers in a cell geometry tree, and ``HalfSpace.__iter__`` to traverse geometry leaves (:issue:`737`).
+* Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
+* Added :class:`~montepy.comments.CommentCollection`, now returned by ``comments`` and ``leading_comments``, so an object can be found by its comments with e.g. ``"keyword" in cell.comments``, or searched by regular expression with ``cell.comments.search`` (:issue:`185`).
+
+**Bugs Fixed**
+
+* Fixed ``leading_comments`` crashing when set with more than one comment (:pull:`972`).
 
 1.4 releases
 ============
 
 #Next Version#
 --------------
-
-**Feature Added**
-
-* Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
-* Added :class:`~montepy.comments.CommentCollection`, now returned by ``comments`` and ``leading_comments``, so an object can be found by its comments with e.g. ``"keyword" in cell.comments``, or searched by regular expression with ``cell.comments.search`` (:issue:`185`).
 
 **Bugs Fixed**
 
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -15,6 +15,7 @@ MontePy Changelog
 **Feature Added**
 
 * Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
+* Added :class:`~montepy.comments.CommentCollection`, now returned by ``comments`` and ``leading_comments``, so an object can be found by its comments with e.g. ``"keyword" in cell.comments``, or searched by regular expression with ``cell.comments.search`` (:issue:`185`).
 
 **Bugs Fixed**
 
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -12,6 +12,10 @@ MontePy Changelog
 #Next Version#
 --------------
 
+**Feature Added**
+
+* Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
+
 **Bugs Fixed**
 
 * Fixed bug where values for user generated objects would be rounded off at five digits no matter what. This will now round off at 15 digits if necessary (:issue:`962`).
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -12,12 +12,6 @@ Next Release
 **Features Added**
 
 * Added ``HalfSpace.replace`` to swap dividers in a cell geometry tree, and ``HalfSpace.__iter__`` to traverse geometry leaves (:issue:`737`).
-* Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
-* Added :class:`~montepy.comments.CommentCollection`, now returned by ``comments`` and ``leading_comments``, so an object can be found by its comments with e.g. ``"keyword" in cell.comments``, or searched by regular expression with ``cell.comments.search`` (:issue:`185`).
-
-**Bugs Fixed**
-
-* Fixed ``leading_comments`` crashing when set with more than one comment (:pull:`972`).
 
 1.4 releases
 ============
@@ -25,8 +19,14 @@ Next Release
 #Next Version#
 --------------
 
+**Feature Added**
+
+* Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
+* Added :class:`~montepy.comments.CommentCollection`, now returned by ``comments`` and ``leading_comments``, so an object can be found by its comments with e.g. ``"keyword" in cell.comments``, or searched by regular expression with ``cell.comments.search`` (:issue:`185`).
+
 **Bugs Fixed**
 
+* Fixed ``leading_comments`` crashing when set with more than one comment (:pull:`972`).
 * Fixed bug where values for user generated objects would be rounded off at five digits no matter what. This will now round off at 15 digits if necessary (:issue:`962`).
 
 
--- a/montepy/comments.py
+++ b/montepy/comments.py
@@ -1,17 +1,19 @@
 # Copyright 2026, Battelle Energy Alliance, LLC All Rights Reserved.
+from collections.abc import Sequence
 import re
 
 from montepy.input_parser.syntax_node import CommentNode
 
 
-class CommentCollection(list):
-    """A list of the comments in an object that supports searching by text.
+class CommentCollection(Sequence):
+    """A read-only collection of the comments in an object that supports searching by text.
 
-    This is a :class:`list` of :class:`~montepy.input_parser.syntax_node.CommentNode`
-    instances, and behaves like a normal list in every way. In addition,
-    checking a string with the ``in`` operator searches the *text* of all
-    comments, so an object can be found by its comments; anything other than a
-    string keeps the normal list membership behavior.
+    This is a :class:`~collections.abc.Sequence` of
+    :class:`~montepy.input_parser.syntax_node.CommentNode` instances, so it
+    supports indexing, slicing, iteration, and ``len()``. Checking a string
+    with the ``in`` operator searches the *text* of all comments, so an
+    object can be found by its comments; anything other than a string keeps
+    normal membership behavior.
 
     Examples
     --------
@@ -24,13 +26,34 @@ class CommentCollection(list):
         True
         >>> problem.materials[1].comments.search(r"(?i)LIGHT", regex=True)[0].contents
         'light water'
+
+    Parameters
+    ----------
+    comments : iterable of CommentNode
+        the comments in this collection.
     """
 
+    __slots__ = ("_comments",)
+
+    def __init__(self, comments=()):
+        self._comments = list(comments)
+
+    def __getitem__(self, i):
+        if isinstance(i, slice):
+            return CommentCollection(self._comments[i])
+        return self._comments[i]
+
+    def __len__(self):
+        return len(self._comments)
+
     def __contains__(self, item):
-        """For a string: search the text of all comments; otherwise: normal list membership."""
+        """For a string: search the text of all comments; otherwise: normal membership."""
         if isinstance(item, str):
-            return bool(self.search(item))
-        return super().__contains__(item)
+            return any(item in comment for comment in self._comments)
+        return item in self._comments
+
+    def __repr__(self):
+        return f"CommentCollection({self._comments!r})"
 
     def search(self, pattern, regex=False):
         """Searches the text of the comments in this collection.
@@ -71,4 +94,4 @@ class CommentCollection(list):
             raise TypeError(
                 f"pattern must be a str, or a pattern compiled from a str. {pattern} given."
             )
-        return CommentCollection(c for c in self if matcher(c.contents))
+        return CommentCollection(c for c in self._comments if matcher(c.contents))
--- a/montepy/comments.py
+++ b/montepy/comments.py
@@ -8,70 +8,26 @@ class CommentCollection(list):
     """A list of the comments in an object that supports searching by text.
 
     This is a :class:`list` of :class:`~montepy.input_parser.syntax_node.CommentNode`
-    instances, and behaves like a normal list in every way. In addition, it
-    supports searching the text of its comments.
+    instances, and behaves like a normal list in every way. In addition,
+    checking a string with the ``in`` operator searches the *text* of all
+    comments, so an object can be found by its comments; anything other than a
+    string keeps the normal list membership behavior.
 
     Examples
     --------
 
-    Searching comments with ``in``
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    .. doctest::
 
-    Checking a string with the ``in`` operator searches the *text* of all
-    comments in the collection, so an object can be found by its comments:
-
-    .. testcode::
-
-        import montepy
-
-        problem = montepy.read_input("foo.imcnp")
-
-        for material in problem.materials:
-            if "light water" in material.comments:
-                print("found material:", material.number)
-
-    .. testoutput::
-
-        found material: 1
-
-    Checking anything other than a string keeps the normal :class:`list`
-    behavior, so membership tests for
-    :class:`~montepy.input_parser.syntax_node.CommentNode` instances are
-    unchanged.
-
-    Searching comments by pattern
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-    :func:`search` finds the comments matching a substring, or a regular
-    expression with ``regex=True``:
-
-    .. testcode::
-
-        material = problem.materials[1]
-        for comment in material.comments.search(r"(?i)LIGHT", regex=True):
-            print(comment.contents)
-
-    .. testoutput::
-
-        light water
-
+        >>> import montepy
+        >>> problem = montepy.read_input("foo.imcnp")
+        >>> "light water" in problem.materials[1].comments
+        True
+        >>> problem.materials[1].comments.search(r"(?i)LIGHT", regex=True)[0].contents
+        'light water'
     """
 
     def __contains__(self, item):
-        """Checks if a string is in the text of any comment, or if a comment is in this list.
-
-        Parameters
-        ----------
-        item : str or object
-            a string to search the text of all comments for, or any other
-            object to check list membership for.
-
-        Returns
-        -------
-        bool
-            for a string: True iff the string is contained in any comment's
-            ``contents``; otherwise: normal list membership.
-        """
+        """For a string: search the text of all comments; otherwise: normal list membership."""
         if isinstance(item, str):
             return bool(self.search(item))
         return super().__contains__(item)
@@ -102,25 +58,17 @@ class CommentCollection(list):
         TypeError
             if ``pattern`` is not a str, or a pattern compiled from a str.
         """
-        if isinstance(pattern, re.Pattern):
-            # bytes patterns cannot search the comments' str contents
-            if not isinstance(pattern.pattern, str):
-                raise TypeError(
-                    f"pattern must be a str, or a pattern compiled from a str. {pattern} given."
-                )
+        if isinstance(pattern, re.Pattern) and isinstance(pattern.pattern, str):
             matcher = pattern.search
+        elif isinstance(pattern, str) and regex:
+            matcher = re.compile(pattern).search
+        elif isinstance(pattern, str):
+
+            def matcher(contents):
+                return pattern in contents
+
         else:
-            if not isinstance(pattern, str):
-                raise TypeError(
-                    f"pattern must be a str, or a pattern compiled from a str. {pattern} given."
-                )
-            if regex:
-                matcher = re.compile(pattern).search
-            else:
-
-                def matcher(contents):
-                    return pattern in contents
-
-        return CommentCollection(
-            comment for comment in self if matcher(comment.contents)
-        )
+            raise TypeError(
+                f"pattern must be a str, or a pattern compiled from a str. {pattern} given."
+            )
+        return CommentCollection(c for c in self if matcher(c.contents))
--- a/montepy/comments.py
+++ b/montepy/comments.py
@@ -1,14 +1,14 @@
 # Copyright 2026, Battelle Energy Alliance, LLC All Rights Reserved.
-from collections.abc import Sequence
+from collections.abc import Collection
 import re
 
 
-class CommentCollection(Sequence):
+class CommentCollection(Collection):
     """A read-only collection of the comments in an object that supports searching by text.
 
-    This is a :class:`~collections.abc.Sequence` of
-    :class:`~montepy.input_parser.syntax_node.CommentNode` instances, so it
-    supports indexing, slicing, iteration, and ``len()``. Checking a string
+    This is a :class:`~collections.abc.Collection` of
+    :class:`~montepy.input_parser.syntax_node.CommentNode` instances that
+    also supports indexing, slicing, and ``len()``. Checking a string
     with the ``in`` operator searches the *text* of all comments, so an
     object can be found by its comments; anything other than a string keeps
     normal membership behavior.
@@ -28,8 +28,9 @@ class CommentCollection(Sequence):
 
     Parameters
     ----------
-    comments : iterable of CommentNode
-        the comments in this collection.
+    comments : collections.abc.Iterable
+        the comments in this collection, as an iterable of
+        :class:`~montepy.input_parser.syntax_node.CommentNode`.
     """
 
     __slots__ = ("_comments",)
@@ -41,6 +42,9 @@ class CommentCollection(Sequence):
         if isinstance(i, slice):
             return CommentCollection(self._comments[i])
         return self._comments[i]
+
+    def __iter__(self):
+        return iter(self._comments)
 
     def __len__(self):
         return len(self._comments)
--- a/montepy/comments.py
+++ b/montepy/comments.py
@@ -13,6 +13,8 @@ class CommentCollection(Collection):
     object can be found by its comments; anything other than a string keeps
     normal membership behavior.
 
+    .. versionadded:: 1.5.0
+
     Examples
     --------
 
@@ -64,6 +66,8 @@ class CommentCollection(Collection):
         The search is run against each comment's
         :attr:`~montepy.input_parser.syntax_node.CommentNode.contents`, that
         is the comment's text without its delimiters (``c``/``$``).
+
+        .. versionadded:: 1.5.0
 
         Parameters
         ----------
--- a/montepy/comments.py
+++ b/montepy/comments.py
@@ -2,8 +2,6 @@
 from collections.abc import Sequence
 import re
 
-from montepy.input_parser.syntax_node import CommentNode
-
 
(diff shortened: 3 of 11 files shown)


montepy/input_parser/syntax_node.py
```diff
@@ -937,6 +937,24 @@ def __repr__(self):
...
+        value : str
+            the string to search for.
+
+        Returns
+        -------
+        bool
+            True iff ``value`` is a substring of this comment's
+            ``contents``.
+        return value in self.contents
+
```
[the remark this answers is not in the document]
yrasool: Thanks , this is a copilot reiew so  I looked into this one and I believe it's a false positive. The triple-quoted docstring is closed on the line directly above `return`, so `return value in self.contents` is a normal statement, not part of the docstring. The module imports cleanly and the full test suite (including the new `test_comment_contains`) passes in CI, which would not be possible if there were a syntax/indentation error here.

MicahGale:
@yrasool The formatting in the PR description is a bit off. Could you go back to the template and make sure you have all parts and formatting included?

montepy/input_parser/syntax_node.py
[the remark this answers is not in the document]
MicahGale: Yes it is a false positive.

MicahGale (changes requested):
@yrasool this is change is very well done for what has been implemented so far. A few notes though.

1. docstrings for `__contains__` is hidden by default. While we can expose that for many users that may lead to further confusion. I usually document this behavior at the class level in example code. See `montepy.Material` for an example of this.

2. The feature request wasn't fully fleshed out, and so the exact desired behavior isn't fully defined for you to implement. I'm going to add some comments in the original issue to continue the design discussion.

yrasool:
Thanks for the review @MicahGale!

**Docs:** moved the `in` documentation up to the `CommentNode` class docstring as a runnable `Examples` block, following the `montepy.Material` pattern you pointed to.

**Design:** following the discussion on #185, I implemented the direction you and @tjlaboss converged on:

- New `CommentCollection` class — a `list` subclass, so all existing list behavior (indexing, iteration, equality, `CommentNode` membership) is unchanged, which I hope addresses the breaking-change concern.
- `MCNP_Object.comments` and `leading_comments` now return it.
- `__contains__` calls the search function non-ambiguously: a `str` operand searches the text of all comments; anything else keeps normal list membership. So `if "foo" in cell.comments` now works.
- `search(pattern, regex=False)` handles substring, regex, and compiled patterns, and returns a new `CommentCollection`, so scope selection falls out of which property you search (`comments` vs `leading_comments`).

Two things I deliberately left for the #185 discussion rather than guessing: `NumberedObjectCollection.find_by_comment`/`search_by_comment` (happy to add once you settle the name/semantics), and `trailing_comment` is untouched.

One subtle behavior note for review: previously a `str` could match via `CommentNode.__eq__` only if it equaled the comment's full `str()` rendering (delimiter included). With search semantics that exact-equality path is replaced by substring-of-contents, which I believe is the intent — but flagging it since it is technically a behavior change.

Also merged `develop` in to bring the branch up to date.

montepy/mcnp_object.py
```diff
@@ -296,29 +297,32 @@ def _flush_line_expansion_warning(self, lines, ws):
...
         and any comments that are inside this card.
 
         Returns
         -------
-        list
-            a list of the comments associated with this comment.
+        CommentCollection
+            a list of the comments associated with this comment, which also
+            supports searching the comments' text (e.g. ``"foo" in
+            obj.comments``).
```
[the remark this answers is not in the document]
yrasool: Good catch — fixed in be278f3 ("associated with this object").

montepy/comments.py
```diff
@@ -0,0 +1,113 @@
...
+        CommentCollection
+            a new collection of the comments that matched.
+        """
+        if isinstance(pattern, re.Pattern):
+            matcher = pattern.search
+        elif regex:
+            matcher = re.compile(pattern).search
+        else:
+            if not isinstance(pattern, str):
+                raise TypeError(f"pattern must be a str. {pattern} given.")
```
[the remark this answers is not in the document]
yrasool: Verified and fixed in be278f3: `search()` now raises a consistent `TypeError` for non-str patterns (including with `regex=True`) and for patterns compiled from bytes, which previously failed mid-search with a cryptic error. Added test cases for all three rejection paths and a `Raises` section to the docstring.

tjlaboss:
I don't think `CommentCollection` should subclass `list`. It should be its own type of Collection.

yrasool:
@tjlaboss done in 9d4145b : `CommentCollection` is now its own read-only collection type (`collections.abc.Sequence`), not a `list` subclass, in line with how the project's other collections are their own classes. It keeps indexing, slicing, iteration, and `len()`; slices return a `CommentCollection`. This is arguably more correct than the old plain `list` too: `comments` always returned a fresh copy, so mutating it never wrote back to the problem , the Sequence makes that explicit.

montepy/comments.py
```diff
@@ -0,0 +1,74 @@
...
+    --------
+
+    .. doctest::
+
+        >>> import montepy
+        >>> problem = montepy.read_input("foo.imcnp")
+        >>> "light water" in problem.materials[1].comments
+        True
+        >>> problem.materials[1].comments.search(r"(?i)LIGHT", regex=True)[0].contents
+        'light water'
```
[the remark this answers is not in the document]
yrasool: I believe this is a false positive: `doc/foo.imcnp` is a tracked file in the repository (a symlink to `doc/source/foo.imcnp`), and `make doctest` resolves it — `montepy.Material`'s class docstring on `develop` loads the exact same `read_input("foo.imcnp")` (montepy/data_inputs/material.py:164) and doc-test CI passes. The suggested `tests/inputs/test.imcnp` path would actually fail, since it does not resolve from the doctest working directory.

montepy/comments.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    def __contains__(self, item):
+        """For a string: search the text of all comments; otherwise: normal list membership."""
+        if isinstance(item, str):
+            return bool(self.search(item))
+        return super().__contains__(item)
```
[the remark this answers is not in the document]
yrasool: Fixed in 9d4145b — `__contains__` now short-circuits with `any(item in comment for comment in ...)` instead of building a full result collection. (Same substring semantics, delegating to `CommentNode.__contains__`.)

montepy/input_parser/syntax_node.py
```diff
@@ -937,6 +937,25 @@ def __repr__(self):
...
+        ----------
+        value : str
+            the string to search for.
+
+        Returns
+        -------
+        bool
+            True iff ``value`` is a substring of this comment's
+            :func:`contents`.
+        """
```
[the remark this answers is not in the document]
yrasool: Addressed — the docstring now uses an inline ``contents`` literal here, and `:attr:` (the property role) where it is cross-referenced from `CommentCollection`.

MicahGale:
> Two things I deliberately left for the https://github.com/idaholab/MontePy/issues/185 discussion rather than guessing: NumberedObjectCollection.find_by_comment/search_by_comment (happy to add once you settle the name/semantics), and trailing_comment is untouched.

That was a good idea. I think that's a separate issue (#982), and you can open a PR later for that if you like.

doc/source/api/modules.rst
```diff
@@ -33,6 +33,7 @@ Collections
    :template: myclass.rst
 
    montepy.Cells
+   montepy.CommentCollection
```
MicahGale, line 36: This isn't a truly "user" facing class. I will think about this more.

montepy/input_parser/syntax_node.py
```diff
@@ -825,6 +825,18 @@ class CommentNode(SyntaxNodeBase):
     ----------
     input : sly.lex.Token
         the token from the lexer
+
+    Examples
+    --------
+
+    The ``in`` operator searches the comment's text — its ``contents``
+    without the ``c``/``$`` delimiter:
```
MicahGale: Not an appropriate use of em dash. It should be matched here. I think commas would make more sense:

``` suggestion
    The ``in`` operator searches the comment's text, its ``contents``,
    without the ``c``/``$`` delimiter:
```

Also an example needs to include a code snippet.

montepy/comments.py
```diff
@@ -0,0 +1,97 @@
...
+    Parameters
+    ----------
+    comments : iterable of CommentNode
+        the comments in this collection.
+    """
+
+    __slots__ = ("_comments",)
+
+    def __init__(self, comments=()):
+        self._comments = list(comments)
```
MicahGale: The more I think about it, the contents should be immutable, and I think `tuple` would be the easiest way to do this. Is this your objection to subclassing `list` @tjlaboss?

montepy/comments.py
```diff
@@ -0,0 +1,97 @@
...
+            return CommentCollection(self._comments[i])
+        return self._comments[i]
+
+    def __len__(self):
+        return len(self._comments)
+
+    def __contains__(self, item):
+        """For a string: search the text of all comments; otherwise: normal membership."""
+        if isinstance(item, str):
+            return any(item in comment for comment in self._comments)
```
MicahGale, line 60:
``` suggestion
            return any((item in comment for comment in self._comments))
```

This would make this into a generator expression allowing this to return on the first match.

montepy/comments.py
```diff
@@ -0,0 +1,97 @@
...
+        return item in self._comments
+
+    def __repr__(self):
+        return f"CommentCollection({self._comments!r})"
+
+    def search(self, pattern, regex=False):
```
MicahGale: I think it would be fine to accept `pattern` as either a string or [`re.Pattern`](https://docs.python.org/3/library/re.html#re.Pattern), and drop the `regex` arg. I figure if a user wants to use regular expressions, they can also easily run `re.compile` themselves first. Thoughts @tjlaboss?

montepy/comments.py
```diff
@@ -0,0 +1,97 @@
...
+        elif isinstance(pattern, str):
+
+            def matcher(contents):
+                return pattern in contents
+
+        else:
+            raise TypeError(
+                f"pattern must be a str, or a pattern compiled from a str. {pattern} given."
+            )
+        return CommentCollection(c for c in self._comments if matcher(c.contents))
```
MicahGale, line 106: Also should use a generator expression here.

montepy/comments.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        if isinstance(pattern, re.Pattern) and isinstance(pattern.pattern, str):
+            matcher = pattern.search
+        elif isinstance(pattern, str) and regex:
+            matcher = re.compile(pattern).search
+        elif isinstance(pattern, str):
+
+            def matcher(contents):
+                return pattern in contents
+
```
MicahGale, line 101: Please remove the extra blank lines, it makes it a little harder to read with all the branching.

montepy/comments.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2026, Battelle Energy Alliance, LLC All Rights Reserved.
+from collections.abc import Sequence
+import re
+
+from montepy.input_parser.syntax_node import CommentNode
+
+
+class CommentCollection(Sequence):
```
MicahGale, line 9: I can't decide if this should be a `_Private` class or not. I think users will see it in a similar way to `montepy.Cells` so I think it's fine as is? @tjlaboss

MicahGale (changes requested):
Great improvement @yrasool. A few tweaks. I did leave a few open-ended design questions I want to have a discussion about; please hold off on tackling those until we get a bit more consensus on the design.

montepy/comments.py
tjlaboss: Yeah probably fine

tjlaboss:
`CommentCollection` should subclass `Collection`, not `Sequence`.

montepy/comments.py
tjlaboss: Either it should be mutable and modify the parent object, or it should be immutable and not support operands. A `Collection` needs `__contains__`, `__iter__`, and `__len__`, and is the best fit. 

Subclassing `tuple` would be better than `list`, but it still opens the door to something like

```python
>>> coms = my_nifty_cell.comments
>>> coms += ('foo', 'bar')
>>> type(coms)
<class 'tuple'>
```

montepy/comments.py
tjlaboss: Parentheses around generator expressions are optional when it's the only argument, thus the parens are redundant.

montepy/comments.py
tjlaboss: Yeah, there's no need for the user to have to specify `regex`. The user should be encouraged to make `pattern` a compiled regex if things like `re.IGNORECASE` are desired.

montepy/comments.py
tjlaboss: is generator already

yrasool:
Thanks both! Tweaks are in (0d2605d):

- `regex` arg is gone. `search(pattern)` takes a substring or a compiled `re.Pattern`, and the docstring points at `re.compile(pattern, re.IGNORECASE)` for the case-insensitive case. Nice side effect: the substring branch just compiles an escaped pattern now, so the nested `def` and its blank lines went away too.
- Storage is a `tuple` now, so it's immutable throughout. The class itself is still its own type rather than a tuple subclass, so `coms += (...)` stays a no-go.
- Comma fix applied. On the snippet note, there's a runnable doctest right under that sentence already, but I can expand it if you'd like.

I've left the `modules.rst` question alone until you decide. And thanks for opening #982, I'd be glad to pick that one up once the design settles.

yrasool:
Took care of the CI failures in 99d37d0:

- `test_leading_comments` was a real regression on my end: the `leading_comments` setter only took `(list, tuple, CommentNode)`, so assigning what the getter itself now returns (like `cell.leading_comments = other.leading_comments[0:1]`) blew up. The setter accepts a `CommentCollection` now. Full suite is green locally apart from the pre-existing Windows-only path and line-ending failures.
- `doc-build` turned out to be the `Sequence` base: autodoc pulls inherited members in nitpicky mode, and `Sequence.count`/`index` have C docstrings that don't resolve. So @tjlaboss, CI kind of made the call for us: the base is `collections.abc.Collection` now, with `__getitem__` and slicing implemented directly so indexing still works. The last warning was my `iterable` type reference, swapped for `collections.abc.Iterable`.

doc-test passed, so the `foo.imcnp` doctests are fine.

doc/source/api/modules.rst
yrasool: Sounds good, I'll leave it alone until you decide. Easy to pull it out of `modules.rst` if you'd rather not list it.

montepy/input_parser/syntax_node.py
yrasool: Applied your suggestion in 0d2605d. There's a runnable doctest right under this line, but happy to beef it up if you want a fuller example.

montepy/comments.py
yrasool: Went with a tuple internally in 0d2605d, so it's immutable all the way down. And since the class is its own Collection rather than a tuple subclass, the `coms += (...)` door @tjlaboss mentioned stays closed.

montepy/comments.py
yrasool: Dropped it in 0d2605d. `search(pattern)` now just takes a substring or a compiled `re.Pattern`, and the docstring nudges people toward `re.compile(pattern, re.IGNORECASE)` if they want case-insensitive.

montepy/comments.py
yrasool: Gone in 0d2605d. Once the `regex` arg was dropped I could compile the escaped substring instead of defining a matcher inline, so the nested `def` and the blank lines disappeared with it.

montepy/comments.py
yrasool: Yep, it's already a generator expression (the parens are implied there), so it short-circuits as is. Left it unchanged.

montepy/comments.py
yrasool: Yep, already a generator there.

montepy/comments.py
yrasool: Keeping it public then. It's exported as `montepy.CommentCollection` next to the other collections.

montepy/comments.py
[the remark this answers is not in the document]
MicahGale: Yes. I agree. CI will ultimately show.

montepy/comments.py
MicahGale: Great! Thanks all. TIL.

MicahGale:
> `CommentCollection` should subclass `Collection`, not `Sequence`.

Why @tjlaboss? This is will maintain order and comments are inherently ordered. I think `Sequence` makes the most sense from my basic understanding of `collections.abc`

doc/source/api/modules.rst
MicahGale: I think it makes sense to keep this here the more I think about. I was thinking about demoting it to the end of the list, but breaking alphabetization will just be confusing.

MicahGale:
Other thought I forgot to mention @yrasool all new functions will need a `.. versionadded::` directive. I think `CommentNode` will need a `.. versionchanged::` directive for the `__contains__` behavior.

You can say this was in version 1.5.0

yrasool:
Added the directives in 56ca8bf: `versionadded` on `CommentCollection` and `search()`, `versionchanged` on `CommentNode` for the `in` behavior, and on `comments`/`leading_comments` for the new return type. All marked 1.5.0.

On the Sequence vs Collection question, some context from my side since I made that switch: it was mostly practical. With `Sequence` as the base, doc-build failed because the autosummary template pulls inherited members in nitpicky mode, and `Sequence.count`/`index` carry C docstrings that don't resolve. Order is preserved either way (it stores a tuple, and `__getitem__`/slicing are implemented directly), so the current version still behaves like a sequence in practice. If you two land on `Sequence`, I can just define `count` and `index` explicitly with proper docstrings and that problem goes away. Happy either way.

doc/source/changelog.rst
```diff
@@ -19,6 +19,11 @@ Next Release
 #Next Version#
 --------------
 
+**Feature Added**
+
+* Added support for the ``in`` operator on comments, so a comment's text can be searched with e.g. ``"keyword" in comment`` (:issue:`185`).
+* Added :class:`~montepy.comments.CommentCollection`, now returned by ``comments`` and ``leading_comments``, so an object can be found by its comments with e.g. ``"keyword" in cell.comments``, or searched by regular expression with ``cell.comments.search`` (:issue:`185`).
```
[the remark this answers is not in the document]
MicahGale: False positive.

montepy/mcnp_object.py
```diff
@@ -296,39 +297,49 @@ def _flush_line_expansion_warning(self, lines, ws):
...
-        return list(self._tree["start_pad"].comments)
+        return CommentCollection(self._tree["start_pad"].comments)
 
     @leading_comments.setter
     def leading_comments(self, comments):
-        if not isinstance(comments, (list, tuple, CommentNode)):
+        if not isinstance(comments, (list, tuple, CommentNode, CommentCollection)):
             raise TypeError(
                 f"Comments must be a CommentNode, or a list of Comments. {comments} given."
             )
```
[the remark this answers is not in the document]
yrasool: Addressed in 8892ad9. The setter now names CommentCollection in the accepted inputs message, and the duplicate validation path was removed so invalid elements use the indexed TypeError below.

montepy/mcnp_object.py
```diff
@@ -296,39 +297,49 @@ def _flush_line_expansion_warning(self, lines, ws):
...
         if isinstance(comments, CommentNode):
             comments = [comments]
-        if isinstance(comments, (list, tuple)):
+        if isinstance(comments, (list, tuple, CommentCollection)):
             for comment in comments:
                 if not isinstance(comment, CommentNode):
                     raise TypeError(
```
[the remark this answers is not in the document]
yrasool: Thanks for catching this. Fixed in 8892ad9. The setter now builds the padding nodes with a flat comprehension, and I added test_comments_setter_multiple_comments to cover assigning two comments.

yrasool:
Thanks again for the careful review @MicahGale and @tjlaboss . I pushed 8892ad9 with the two remaining setter fixes: updated the accepted inputs error message and fixed multi-comment assignment, with a focused regression test.

I really appreciated this review cycle. It was fun and engaging, and I would definitely love to keep contributing more and more and learning more from the project.

montepy/comments.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2026, Battelle Energy Alliance, LLC All Rights Reserved.
+from collections.abc import Collection
+import re
+
+
+class CommentCollection(Collection):
```
MicahGale: As discussed let's subclass `Sequence`. As for the docstring issues just use `nitpick_ignore` (or something like that) in `doc/source/conf.py`

montepy/comments.py
```diff
@@ -0,0 +1,96 @@
...
+        CommentCollection
+            a new collection of the comments that matched.
+
+        Raises
+        ------
+        TypeError
+            if ``pattern`` is not a str, or a pattern compiled from a str.
+        """
+        if isinstance(pattern, str):
+            matcher = re.compile(re.escape(pattern)).search
```
MicahGale: Treating the search string as a regular expression is going to have unintended consequences. Just do what you had previously done: define a function which checks `pattern in comment`.

tests/test_integration.py
```diff
@@ -558,6 +558,17 @@ def test_comments_setter(simple_problem):
         cell.leading_comments = 5
 
 
+def test_comments_setter_multiple_comments(simple_problem):
```
MicahGale, line 562: Need to also test `pattern in MCNP_Object.comments` now. I think that would be good in this file.

tests/test_syntax_parsing.py
```diff
@@ -1,6 +1,8 @@
 # Copyright 2024, Battelle Energy Alliance, LLC All Rights Reserved.
 import copy
 from io import StringIO
+import os
```
MicahGale: Why is `os` needed as an import? Generally this file shouldn't test an disk I/O. Also we are moving towards preferring `pathlib.Path` or `os.path`.

tests/test_syntax_parsing.py
```diff
@@ -603,6 +605,69 @@ def test_blank_dollar_comment(self):
...
+        assert isinstance(matches, montepy.CommentCollection)
+        # non-string patterns are rejected
+        with pytest.raises(TypeError):
+            comments.search(5)
+        # bytes patterns cannot search the str contents
+        with pytest.raises(TypeError):
+            comments.search(re.compile(b"fuel"))
+
+
+def test_object_comments_are_collection():
```
MicahGale: This test as written should go in `tests/test_integration.py`

tests/test_syntax_parsing.py
```diff
@@ -603,6 +605,69 @@ def test_blank_dollar_comment(self):
...
+        # strings search the text of all comments
+        assert "fuel" in comments
+        assert "notes" in comments
+        assert "graphite" not in comments
+        # anything else keeps normal list membership behavior
+        assert comments[0] in comments
+        assert syntax_node.CommentNode("c not in there") not in comments
+        assert 5 not in comments
+
+    def test_comment_collection_search(self):
```
MicahGale, line 637: Good test. Please also explicitly test all other functions of `CommentCollection`

montepy/comments.py
```diff
@@ -0,0 +1,96 @@
...
+    def __contains__(self, item):
+        """For a string: search the text of all comments; otherwise: normal membership."""
+        if isinstance(item, str):
+            return any(item in comment for comment in self._comments)
+        return item in self._comments
+
+    def __repr__(self):
+        return f"CommentCollection({self._comments!r})"
+
+    def search(self, pattern):
```
MicahGale: It's not explicit in the dev docs, but we are going towards writing type hints for all our functions. For these new functions please add type hints as needed.

montepy/input_parser/syntax_node.py
```diff
@@ -937,6 +953,24 @@ def __repr__(self):
     def __eq__(self, other):
         return str(self) == str(other)
 
+    def __contains__(self, value):
+        """Checks if a string is found in the contents of this comment.
+
+        This allows searching a comment by its text, e.g.
+        ``"important" in comment``.
+
```
MicahGale, line 961: Add a `.. versionadded::`

montepy/input_parser/syntax_node.py
MicahGale: Looks good.

montepy/comments.py
MicahGale: And I checked with switching to `Sequence` `__add__`, etc. are not implemented. Only: `__iter__`, `__getitem__`, `__len__`

MicahGale (changes requested):
Just a few more comments. 

In addition Coveralls is having some issues, but it does look like some of the changes are not fully covered yet: https://coveralls.io/jobs/184390088.

> I really appreciated this review cycle. It was fun and engaging, and I would definitely love to keep contributing more and more and learning more from the project.

It's been great having your contribution, and we would always be happy to accept more :).

montepy/comments.py
yrasool: Done in 47bdcbd, it's back on `Sequence` with a `nitpick_ignore_regex` entry in `conf.py` for the inherited `count`/`index` return annotations. And thanks for checking `__add__`, that closes the operand worry nicely.

montepy/comments.py
yrasool: Fair point, reverted in 47bdcbd. Substring search is a plain `pattern in contents` function again, and I added a test that regex metacharacters like `a+b` match literally.

tests/test_integration.py
yrasool: Added in 47bdcbd, `test_object_comments_are_collection` now lives in `test_integration.py` and checks both a plain string and a compiled pattern against `cell.comments`.

tests/test_syntax_parsing.py
yrasool: It was only there for the test that read an input file directly. That test moved to `test_integration.py` and uses the `simple_problem` fixture now, so the `os` import is gone.

tests/test_syntax_parsing.py
yrasool: Moved in 47bdcbd, and it uses the `simple_problem` fixture instead of reading the file itself.

tests/test_syntax_parsing.py
yrasool: Added `test_comment_collection_sequence` in 47bdcbd, it covers indexing (including negative), slicing returning a `CommentCollection`, iteration, `reversed`, `index`, `count`, `len`, and `repr`. That should also close most of the coveralls gaps.

montepy/comments.py
yrasool: Added in 47bdcbd, all the new functions have hints now, e.g. `search(self, pattern: str | re.Pattern) -> CommentCollection`.

montepy/input_parser/syntax_node.py
yrasool: Added in 47bdcbd.

MicahGale (approved):
This looks good. Thanks for working through the details with us @yrasool.

On a side note: were the copilot reviews helpful to you? I don't like that it uses the credits from the user, and sometime just doesn't work because of that.

yrasool:
> This looks good. Thanks for working through the details with us @yrasool.
> 
> On a side note: were the copilot reviews helpful to you? I don't like that it uses the credits from the user, and sometime just doesn't work because of that.

You're welcome. I would say yes for pointing out obvious mistakes yes Copilot is helpful . However if it was not for detailed reviews from both you and @tjlaboss it would be a little difficult to exactly identify and fix the code and learn.

PR #1003, opened 2026-08-03
Fix docstrings naming parameters that are not in the signature

### LINKED ISSUE #1002: Four docstrings document parameter names that are not in the signature
Filing this first per the developer's guide ("Tie your work to an issue"). I have the fix ready and will open a PR against `develop` referencing this issue.

Four docstrings name a parameter that is not in the function's signature. All four are rendered by Sphinx, so the published API docs carry the wrong names.

**1. `NumberedObjectCollection.symmetric_difference` and `.symmetric_difference_update`** (`montepy/numbered_object_collection.py`)

Both take a single `other`, but document `others : Self / the other collections to compare to.` — the exact text used by `difference(*others)` and `difference_update(*others)` a few lines above, which really are varargs. Symmetric difference is binary, so the plural cannot be right here.

**2. `PaddingNode.append` and `ListNode.append`** (`montepy/input_parser/syntax_node.py`)

Both are `append(self, val, ...)` and both document `node`.

**3. `Nuclide._parse_fancy_name`** (`montepy/data_inputs/nuclide.py`)

The Parameters section has two entries for one argument — a bare `identifier` with no type or description, immediately followed by `idenitifer : Union[str, int, element, Nucleus, Nuclide]` with the letters transposed.

**4. `Surface.update_pointers`** (`montepy/surfaces/surface.py`)

Signature is `update_pointers(self, surfaces, data_inputs)`, docstring says `data_cards`. This looks like a leftover from the card → input rename: `Material.update_pointers` and `ThermalScatteringLaw.update_pointers` both document `data_inputs` correctly.

Docstrings only, no behaviour change, so there is nothing to add a test for — happy to be told otherwise.

karpovantonme:
### Description

Fixes #1002.

Four docstrings name a parameter the function does not take. Details and the reasoning for each are in the issue; in short:

- `symmetric_difference` / `symmetric_difference_update` document `others` (plural, "the other collections") but take a single `other` — the text was copied from `difference(*others)` above, which really is varargs.
- `PaddingNode.append` and `ListNode.append` document `node`; both signatures are `append(self, val, ...)`.
- `Nuclide._parse_fancy_name` has two Parameters entries for one argument: a bare `identifier`, then `idenitifer` with the letters transposed.
- `Surface.update_pointers` documents `data_cards`; the signature is `data_inputs`, and `Material.update_pointers` and `ThermalScatteringLaw.update_pointers` already document it that way.

Docstrings only — no code, no behaviour change, so there is nothing for a test to assert. Say the word if you would rather have it covered some other way.

---

### General Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable) — not applicable, docstrings only.

---

### LLM Disclosure

1. Are you?

   - [x] A human user

  - [x] Yes
      - Model(s) used: Claude Opus (Claude Code)
  - [ ] No

The four were found by comparing every NumPy-style `Parameters` block in the package against the actual signature, then reading each hit to rule out decorators, `**kwargs` and the property-setter documentation style used in `fill.py` and `importance.py` (those two are deliberate and were left alone).

<details open>

<summary><h2>Documentation Checklist</h2></summary>

- [x] I have added [type hints](https://peps.python.org/pep-0484/) to all functions as needed.
- [ ] I have marked all changes with the `.. versionchanged::` or `.. versionadded::` directives — not applicable, no API change.

</details>

----
📚 Documentation preview 📚: https://montepy--1003.org.readthedocs.build/en/1003/

----
📚 Documentation preview 📚: https://montepy--1003.org.readthedocs.build/en/1003/

### DIFF
--- a/montepy/data_inputs/nuclide.py
+++ b/montepy/data_inputs/nuclide.py
@@ -760,8 +760,8 @@ class Nuclide:
 
         Parameters
         ----------
-        identifier
-        idenitifer : Union[str, int, element, Nucleus, Nuclide]
+        identifier : Union[str, int, element, Nucleus, Nuclide]
+            the name to parse.
 
         Returns
         -------
--- a/montepy/input_parser/syntax_node.py
+++ b/montepy/input_parser/syntax_node.py
@@ -698,8 +698,8 @@ class PaddingNode(SyntaxNodeBase):
 
         Parameters
         ----------
-        node : str, CommentNode
-            node
+        val : str, CommentNode
+            the node to append.
         is_comment : bool
             whether or not the node is a comment.
         """
@@ -1862,8 +1862,8 @@ class ListNode(SyntaxNodeBase):
 
         Parameters
         ----------
-        node : ValueNode, ShortcutNode
-            node
+        val : ValueNode, ShortcutNode
+            the node to append.
         from_parsing : bool
             If this is being append from the parsers, and not elsewhere.
         """
--- a/montepy/numbered_object_collection.py
+++ b/montepy/numbered_object_collection.py
@@ -1068,8 +1068,8 @@ class NumberedObjectCollection(ABC):
 
         Parameters
         ----------
-        others : Self
-            the other collections to compare to.
+        other : Self
+            the other collection to compare to.
 
         Returns
         -------
@@ -1086,8 +1086,8 @@ class NumberedObjectCollection(ABC):
 
         Parameters
         ----------
-        others : Self
-            the other collections to compare to.
+        other : Self
+            the other collection to compare to.
         """
         self ^= other
         return self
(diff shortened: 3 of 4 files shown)


MicahGale (commented):
The changes look good for the most part. Though, @karpovantonme why did you remove your disclosure of using Claude code?

karpovantonme:
Sorry — that was carelessness with the form on my part, not an attempt to hide anything. I changed the first box to say I'm a human, since I am the one submitting this and standing behind it, and flipped the second one in the same edit without thinking it through. I've put it back.

To be explicit: yes, Claude Code was used. It ran the sweep that compares every numpydoc `Parameters` block in the package against the actual signature, and it wrote the patch. I read the diff before submitting and can explain each of the four changes — including why `fill.py` and `importance.py` are *not* in it, since documenting the setter's parameter in the property's docstring is deliberate there and Sphinx renders it from the getter.

tjlaboss:
The changelog test failure can be ignored in this case.

MicahGale (approved):
Looks good. I agree we can ignore the changelog requirement. Thanks @karpovantonme!

